Issue #1, opened 2026-01-21
AtomicArc implementation is unsound (use-after-free under concurrency, caught by Miri)


biryukovmaxim:
# `AtomicArc` is unsound: use-after-free under concurrent load/store (caught by Miri)

## Summary

The current `AtomicArc<T>` implementation based on `AtomicPtr<T>` is **unsound**.
Under concurrent `load()` / `store()`, it can trigger **use-after-free** and
**undefined behavior**, which is reliably detected by Miri.

This is a fundamental lifetime issue and cannot be fixed by adjusting atomic
orderings.

---

### Test added

```rust
#[cfg(test)]
mod tests {
    use super::*;

    #[test]
    fn test_concurrent_load_store() {
        use std::thread;

        let atomic = Arc::new(AtomicArc::new(Arc::new(42)));
        let mut handles = vec![];

        // Spawn multiple readers
        for _ in 0..10 {
            let atomic = atomic.clone();
            handles.push(thread::spawn(move || {
                for _ in 0..1000 {
                    let value = atomic.load();
                    assert!(*value == 42 || *value == 43);
                }
            }));
        }

        // Spawn writers
        for _ in 0..5 {
            let atomic = atomic.clone();
            handles.push(thread::spawn(move || {
                for _ in 0..1000 {
                    atomic.store(Arc::new(43));
                }
            }));
        }

        for h in handles {
            h.join().unwrap();
        }
    }
}
```
### Run with Miri
```bash
cargo +nightly miri test -p vprogs-core-atomics
```

###Result
`Undefined Behavior: in-bounds pointer arithmetic failed:
alloc74184 has been freed, so this pointer is dangling`

I'm also wondering why not use well tested arc-swap which may be used for the same purpose

hmoog:
You are absolutely right and replacing with arc-swap was actually on my to-do list.

It's fixed here: https://github.com/kaspanet/vprogs/pull/2 maybe you can review then we can merge it

PR #2, opened 2026-01-21
Fix: Use arc-swap instead of custom implementation

hmoog:
This PR fixes the issue mentioned in: https://github.com/kaspanet/vprogs/issues/1

----

 Files modified in scheduling/scheduler/src/:                                                                                                                                                                              
                                                                                                                                                                                                                            
  1. resource_access.rs: Replaced AtomicOptionArc with ArcSwapOption:                                                                                                                                                       
    - empty() → empty() (same)                                                                                                                                                                                              
    - new(...) → new(...) (same)                                                                                                                                                                                            
    - load() → load_full()                                                                                                                                                                                                  
    - store(...) → store(...) (same)                                                                                                                                                                                        
    - swap(None) → swap(None) (same)                                                                                                                                                                                        
    - publish(v) → compare_and_swap(&None::<Arc<_>>, Some(v)).is_none()                                                                                                                                                     
  2. runtime_tx.rs: Same pattern as above                                                                                                                                                                                   
  3. state_diff.rs: Same pattern as above                                                                                                                                                                                   
                                                                                                                                                                                                                            
  Files modified in scheduling/scheduler/:                                                                                                                                                                                  
                                                                                                                                                                                                                            
  4. Cargo.toml: Added direct arc-swap = "1.7" dependency                                                                                                                                                                   
                                                                                                                                                                                                                            
  Files modified/deleted in core/atomics/:                                                                                                                                                                                  
                                                                                                                                                                                                                            
  5. src/lib.rs: Removed AtomicArc and AtomicOptionArc exports                                                                                                                                                              
  6. src/atomic_arc.rs: Deleted                                                                                                                                                                                             
  7. src/atomic_option_arc.rs: Deleted                                                                                                                                                                                      
                                                                                                                                                                                                                            
  Documentation:                                                                                                                                                                                                            
                                                                                                                                                                                                                            
  8. core/README.md: Removed AtomicArc from the list of atomics                                                                                                                                                             
                                                                                                                                                                                                                            
  The vprogs-core-atomics crate now only contains AtomicAsyncLatch and AtomicEnum.


biryukovmaxim:
It fixes the issue and looks good 
Imo it makes sense to use `fn load() -> Guard` in cases where it's possible, it's unrelated to the issue so I would only recommend to add todos to investigate where is possible to switch to guard

hmoog:
> It fixes the issue and looks good Imo it makes sense to use `fn load() -> Guard` in cases where it's possible, it's unrelated to the issue so I would only recommend to add todos to investigate where is possible to switch to guard

I just switched to load where it is applicable immediately (no need for todos).

biryukovmaxim (approved):

PR #42, opened 2026-05-07
Runtime poc

biryukovmaxim:
## Summary

Adds `runtime-processor`, a singleton-config runtime program for the RISC-0 backend, together with the covenant/permission-spend hardening it relies on. The program models a small account system — one config resource plus user resources, each guarded by pluggable locks and unlocked by offset-referenced signatures.

## Commits (on top of `master`)

- **add runtime-processor crate with lock/signer trait design** — locks and signers behind generic traits with associated unlocker types; the ix wire format references signatures by offset so one signature can unlock multiple resources; zerocopy DST config layout. Ships Schnorr/Multisig/Unlocked locks, Schnorr-sig and prev-tx-V1 witness signers, and a genesis-gated `Init` action.
- **split AuthContext per Unlocker type and gate image lock** — `Multisig` is its own unlocker type/bucket, so a Schnorr-targeted unlocker cannot satisfy a Multisig lock at the type level; five signer kinds; multisig contributions land aggregated per resource, matchers reject non-strict-ascending/duplicate slices. The forgeable `env::verify` image lock is gated behind `experimental-image-lock` (off by default) pending a real in-guest verifier.
- **add updater_idx to Update/Init actions for positional resource addressing** — resources now arrive in lexicographic-id order, so each action carries a `updater_idx` pointing at its target resource instead of assuming index 0; the bound is rejected at decode time.
- **add user resource kind with transfer + lock rotation; deny panic in guest** — a `kind` byte (config/user), domain-separated user-address derivation, `UserInit`/`Transfer`/`UpdateUserLock` actions; the crate is locked against panicking constructs, since a panicking transaction is un-provable in the guest.
- **Rebuild guest ELFs**
- **Constrain covenant/permission spends and align exit hash domain (#76, #77, #78)** — pin the covenant continuation output value so funds cannot be diverted (#76); constrain the permission payout so an attacker cannot pay dust and pocket the remainder (#77); single-source the permission-tree hash domain so committed exits stay spendable on- and off-chain (#78); dev settlement path gains the two-output exit layout.
- **Switch runtime guest digests to SHA-256 precompile** — resource ids, lock identity hashes, and the Schnorr signed-message digest move from software BLAKE3 to the precompile-backed SHA-256, with a one-byte `Domain` enum for separation.

## Rebased onto master

The branch was reconciled onto current `master`. The settlement, proof-receipt-cache, and superseded-suffix re-aggregation work that used to appear here as duplicate commits has since landed on `master` via #83, #85, and #90 — and the tail plumbing fixes (settlement-observer rename, aggregate-prover `ReceiptStore`, park-wake, `settle_one` outcomes, sim `bundle_size` threading) were absorbed into those merges too. All of it has been dropped, so the branch now carries only the runtime commits above.

## Verification (dev mode)

- `cargo check --workspace --all-targets` — clean
- all five guest ELFs rebuild byte-identically from the reconciled source (`build-guests.sh`)
- `permission_spend` test-suite — 12/12
- `runtime-processor` e2e, which drives the guest ELF through the risc0 executor — 11/11

Real proofs (CUDA) to be run on the proving machine.


zk/backend/risc0/runtime-processor/src/action.rs
```diff
@@ -0,0 +1,229 @@
...
+    auth_ctx: &AuthContext,
+) -> AbiResult<()> {
+    match &action.body {
+        ActionBody::Update { updater_idx, new_min_withdrawal_amount, new_lock } => {
+            apply_update(*updater_idx, *new_min_withdrawal_amount, new_lock, resources, auth_ctx)
+        }
+        ActionBody::Init { updater_idx, new_min_withdrawal_amount, new_lock } => {
+            apply_init(*updater_idx, *new_min_withdrawal_amount, new_lock, resources, auth_ctx)
+        }
+        ActionBody::UserInit { user_idx, initial_balance, initial_lock } => {
```
hmoog, line 33: The UserInit doesn't take or check the auth_ctx.

Is this intended? Couldn't a prover just pick a lock over their own pubkey, address the (new) derived resource, and emit UserInit { initial_balance: u64::MAX } causing the resource to be created fully funded from nothing?

Maybe I am missing something but just wanted to raise this regardless since it looks weird.

hmoog (changes requested):
It's a lot of code and I primarily skimmed through most of it, trying to find logic bugs or obvious mistakes.

I flagged a single issue that looked weird at first glance but let's maybe have another review / hardening round later to not stall progress any longer.

zk/backend/risc0/runtime-processor/src/action.rs
biryukovmaxim: This part indeed not covered. It should add one more check, that tx pays to output that belongs to L2 treasure. It's covered more accurate in the branch called poc-continue

PR #43, opened 2026-05-08
Feat: Skip unsupported tx versions + ABI wire-format refactor

hmoog:
## Summary

Adds a skip path for unsupported L1 transaction versions and tightens the zk-abi wire format along the way. 

### Skip path for unsupported tx versions

- `Inputs` / `InputCommitment` split into a fixed prefix (`version`, `tx_id`, `merge_idx`) plus an `Option<ExecutionInput>` / `Option<ExecutionContext>` that is present iff the version is supported. 
- New `transaction_processor::ErrorCode` (`VersionIncompatible`, `TxIdMismatch`, `MissingExecutionInputs`) emitted in `OutputCommitment::Error`.
- Batch processor enforces the shape invariant: a skipped tx (no `execution_context`) MUST pair with `OutputCommitment::Error(Guest(VersionIncompatible))`; anything else is a forged journal and the prover panics.
- Activity digest is computed for every tx (executed or skipped) so the L1-derived digest still matches.
 
### Wire format

- `Transaction::encode` / `decode` own a length-prefixed blob; the inner payload-and-rest-preimage shape is no longer the caller's concern.
- `ExecutionInput` wire layout: `context_hash | tx(blob) | resource_count | per-resource (header + data)`. Resource entries are now interleaved (header + data, one after another) rather than two separate sections. 
- Resource section count is driven by `access_metadata.len()`; no extra on-wire count prefix. 
- Journal layout simplified: dropped per-entry opcodes, length prefixes, and the `JournalEntry` enum. The tx-processor journal is now `InputCommitment` immediately followed by `OutputCommitment`, decoded as a single struct rather than an enum stream.
- Inputs.batches moved to the last field of the struct; LaneProof::decode signature changed from mut buf: &'a [u8] to buf: &mut &'a [u8] so the outer cursor advances past it. Inputs::encode parameter order mirrors the new struct field order (lane_proof before batches).
 

### Codec helpers

- `Reader::slice_as<T>` — zero-copy slice cast for fixed-layout types, with `count.checked_mul` overflow guard. 
- `Reader::bytes` / `array` rewritten with `split_at_checked` / `split_first_chunk`, eliminating manual length checks.
- New `MutReader` trait + `bytes_mut` / `blob_mut` for mutable cursor reads (used by `Resource::decode` for in-place backing slices).
- `InputResourceCommitment` made `repr(C) FromBytes` and decoded as a zero-copy `& [InputResourceCommitment]` via `slice_as`.
- core_smt::proving::Leaf made repr(C) FromBytes; Proof.leaves / Proof.siblings switched from Vec to zero-copy &[Leaf] / &[[u8; 32]] via slice_as.                                                                          
- batch_processor::StateTransition made repr(C) FromBytes Unaligned with owned fields; decoded via Reader::array_as::<StateTransition> (custom decode and SIZE removed in favor of size_of::<Self>()).
- New batch_processor::Batches<'a> iterator over the batches in Inputs (mirrors TransactionJournals); replaces Vec<Batch<'a>>. Stores (remaining: u32, buf: &[u8]), decodes one batch per next(), fuses on error. first() peeks the head as Result<Batch<'a>> by consuming a Copy of the iterator. Eliminates the per-bundle Vec allocation in the batch processor guest. 

### Verifier cleanup (`batch_processor::Abi`)

- Renamed `verify` → `process_bundle`, `verify_batches` → `process_batches`, `verify_batch` → `process_batch`, `verify_txs` → `verified_activity_digest`, `check_input_resource` → `verified_bundle_idx`. 
- `verified_journal` consolidates per-tx checks: decode, proof verification, shape invariants, strictly-increasing `merge_idx`.
- Per-resource output commitment 1:1 with input resources is now asserted (previously silently allowed missing outputs). 

### Other

- `OutputResourceCommitments` iterator file removed on the input side (replaced by the zero-copy slice).
- Removed unused `wire_size` helpers (`OutputCommitment::wire_size`, `OutputResourceCommitment::wire_size`, `Error::wire_size`) — dead code chain (the outer one was the only caller of the inner two).
- `for_l2_test` now calls `tx.finalize()` after mutating fields, fixing a stale-`tx_id` bug in test setup.


core/smt/src/proving/leaf.rs
```diff
@@ -1,34 +1,23 @@
...
+use vprogs_core_codec::Writer;
+use zerocopy::{FromBytes, Immutable, KnownLayout, little_endian::U16};
 
-use vprogs_core_codec::{Reader, Result};
-
-/// Zero-copy view of a single leaf entry in the proof wire format.
-pub struct Leaf<'a> {
+/// Zero-copy POD view of a single leaf entry in the proof wire format.
+#[repr(C)]
+#[derive(FromBytes, Immutable, KnownLayout)]
```
biryukovmaxim: Can you add here Unaligned? It will ensure we never face alignment errors if structure has higher alignment, and indirectly it also makes no paddings(probably the same as knownlayout requirement

zk/abi/src/batch_processor/abi.rs
```diff
@@ -26,195 +27,240 @@ pub struct Abi<'a> {
...
         let new_seq_commit = this.new_seq_commit(&new_lane_tip, last_blue_score);
         let prev_state = this.inputs.proof.root::<Blake3>().expect("proof root");
-        let new_state = this.inputs.proof.compute_root::<Blake3>(|i| this.latest_hash(i));
+        let new_state = this.inputs.proof.compute_root::<Blake3>(|i| this.latest_value_hash(i));
 
+        // Commit resulting state transition.
         StateTransition::encode(
             journal,
-            (&prev_state, this.inputs.batches[0].prev_lane_tip),
+            (&prev_state, this.inputs.batches.first().unwrap().prev_lane_tip),
```
biryukovmaxim, line 54: Does it assume we have at least single batch/tx?

It must be possible to prove lack of activity from block x to block y. Basically we need ability to update anchor block in cases when we have no activity at all, I think it will help syncing

biryukovmaxim (changes requested):

core/smt/src/proving/leaf.rs
hmoog: Good idea, applied.

Small correction on the rationale though: Unaligned doesn't ensure no alignment errors regardless of struct alignment - it compile-time requires alignment 1 (the derive refuses to compile if the struct has higher alignment). So it's a guard against future field changes that would silently raise alignment, not a fix for higher alignment. The "no padding" property is already enforced by FromBytes / TryFromBytes.

Either way, adding it is worth it for the compile-time invariant - applied to Leaf, AccessType, AccessMetadata, ResourceId, and InputResourceCommitment.

zk/abi/src/batch_processor/abi.rs
hmoog: Good catch. We currently can't - the guest always calls lane_tip_next, but the L1 bridge skips it and carries lane_tip/lane_blue_score forward when a block has no L2 activity.

Today this hasn't surfaced because the scheduler doesn't emit empty batches, but you're right that we need this for syncing/lack-of-activity proofs.

Plan: add an is_empty check on tx_journals in process_batch that returns *batch.prev_lane_tip (matching L1's carry-forward path), and keep last_blue_score at the lane's last-active value across empty batches. A single empty batch is sufficient to advance seq_commit to block y with no activity, so no need to relax the "non-empty bundle" assertion.

Will land in a follow-up since it's orthogonal to this PR.

biryukovmaxim (approved):

PR #46, opened 2026-05-13
Feat: Settle empty batches + further verifier API reorg

hmoog:
**Summary**

Adds settlement for empty batches and rebuilds the batch-processor verifier around a proper type-owned lifecycle.

**Empty-batch settlement**

- New chain assertion in the verifier: each batch's prev_lane_blue_score must match the running carry-forward, mirroring the existing prev_lane_tip chain check.
- Empty batches (no tx_journals) no longer advance the lane state. The verifier carries lane_tip and lane_blue_score forward unchanged when a batch carries no activity, and only advances them when the batch is active.
- Batch.lane_blue_score field renamed to prev_lane_blue_score to reflect what it actually is (lane state entering the batch, not after).
- TransactionJournals::is_empty() added so the verifier can branch on emptiness without iterating the buffer.
- Result: a covenant can be settled with any single empty batch from a quiet lane period, producing a state transition that advances seq_commit without touching the SMT state root or lane tip. Aligns with the protocol's covenant-advance semantics during inactivity.

**Verifier API reorg (batch_processor::Verifier)**

- batch_processor::Abi (unit-struct namespace) deleted. Replaced with batch_processor::Verifier, a proper struct that owns the bundle's decoded inputs, pre-state, and accumulating post-state.
- Public lifecycle: Verifier::new (decode + snapshot pre-state from first batch) → verify_batches (verify chain assertions + return derived (lane_tip, lane_blue_score)) → commit_state_transition (compute roots and emit the settlement journal).
- Encoding logic moved into commit_state_transition. Consumers no longer assemble StateTransition::encode calls themselves; the Verifier owns the full "decode → verify → emit" flow.
- V: Fn bound relaxed to V: FnMut on the journal-verifier closure, so callers can pass stateful closures (instrumentation, counters, debugging) without a workaround.
- Private helpers grouped by which public method they support: verify_activity and chain for verify_batches; prev_root, new_root, new_seq_commit for commit_state_transition.

**Guest orchestration inlined**

- batch_processor::process_bundle free function removed. The batch-processor guest's main.rs now drives Verifier::new → verify_batches → commit_state_transition directly. The three-line process_bundle wrapper was no longer carrying its weight once Verifier became a proper public type.
- transaction_processor::Abi unit struct removed. process_transaction is now a module-level free pub fn.
- risc0-api::Guest unit-struct wrapper removed (same pattern). The transaction-processor guest's main.rs now calls process_transaction(&mut Host, &mut Journal, ...) directly.
- Result: both guest binaries' main.rs files now read as the workflow they actually run.

**Test setup**

- proving_e2e.rs::batch_proof_bundle_of_two and settlement_e2e.rs::batch_proof_bundles_two_batches now also chain metadata_2.prev_lane_blue_score = metadata_1.blue_score, mirroring the existing manual prev_lane_tip chaining. Production does this through the bridge; the K=2 tests bypass the bridge and need to chain manually for the new assertion to pass.

**Build / dev environment**

- .dockerignore patterns globbed to **/target/ / **/target-idea/ so nested workspace target dirs no longer leak into the Docker build context (was ~4 GB; now negligible).
- risc0-guest-builder Docker image bumped from r0.1.88.0 to r0.1.91.1 to match the rustc version (1.91) required by the current kaspa-* dep pins.
- .gitignore covers target-idea (separate target dir for IDE-driven builds; avoids fingerprint thrash with CLI builds in target/).


zk/abi/src/batch_processor/verifier.rs
```diff
@@ -22,47 +22,30 @@ use crate::{
     },
 };
 
-/// Bundle proof workspace exposing per-step helpers for guest orchestration.
-pub struct Abi<'a> {
+/// Verifies a bundle and accumulates its post-state.
+pub struct Verifier<'a, V: Fn(&[u8; 32], &[u8])> {
```
biryukovmaxim: likely FnMut fits also

biryukovmaxim (changes requested):

zk/abi/src/batch_processor/verifier.rs
hmoog: done

zk/abi/src/transaction_processor/abi.rs
```diff
@@ -8,46 +8,41 @@ use crate::{
...
+    // Execute guest closure (if version is supported).
+    let Inputs { version, tx_id, merge_idx, mut execution_input } = inputs;
+    let result = match version {
+        Transaction::V1 => 'v1: {
+            let Some(exec) = execution_input.as_mut() else {
+                break 'v1 Err(ErrorCode::MissingExecutionInputs.into());
+            };
+
+            if tx_id.as_slice() != exec.tx.id() {
+                break 'v1 Err(ErrorCode::TxIdMismatch.into());
```
biryukovmaxim: these two are errors are only possible in case of mischief behavior of host. it makes sense to panic here instead of commitin to the error. error is basically a skipped execution that is possible to prove. such things we should not prove at all

biryukovmaxim (changes requested):

PR #50, opened 2026-05-18
Refactor: Extract Hasher trait into standalone core/hashing crate

hmoog:
**Summary**

Pulls the Hasher trait and the Blake3 impl out of core/smt into a new core/hashing crate. The trait isn't SMT-specific; lifting it lets other tree implementations (the upcoming streaming Merkle tree for permission
commitments, future variants) consume the abstraction without taking a dependency on vprogs_core_smt purely for a generic hash trait.

Also extends the trait with hash_with_domain and hash_parts_with_domain so consumers can express domain-separated hashing without constructing intermediate buffers manually.

**Why**

Today the Hasher trait lives in vprogs_core_smt, but the trait, the Blake3 impl, and the broader concept of "32-byte cryptographic hash with optional domain separation" have nothing to do with sparse Merkle trees. As we
extract additional tree primitives into their own core crates, having every one of them depend on vprogs_core_smt for a generic abstraction inverts the natural dependency direction.

EMPTY_HASH stays in core/smt: the all-zeros sentinel is part of the SMT's subtree-compression scheme (H(empty, empty) → empty), not a hash-function property. Different trees (e.g. the permission tree) use different empty-marker schemes, so this constant doesn't belong in a domain-neutral hashing crate.

**Changes**

New crate core/hashing (no_std):
- Hasher trait:
  - hash(data: impl AsRef<[u8]>) -> [u8; 32] — required.
  - hash_with_domain<const N>(domain: &[u8; N], data: impl AsRef<[u8]>) — default impl, delegates to hash_parts_with_domain with a one-element iterator.
  - hash_parts_with_domain<const N>(domain: &[u8; N], parts: impl IntoIterator<Item = impl AsRef<[u8]>>) — required; streams the domain tag followed by each part through the hasher's incremental API, no intermediate allocation.
- Blake3 impl using blake3::Hasher streaming for hash_parts_with_domain.

core/smt migration:
- Deleted src/hashing/ (trait + impl moved out).
- Moved EMPTY_HASH into src/empty_hash.rs. 
- Rewrote node.rs::hash_internal / hash_leaf to use hash_parts_with_domain, replacing manual [0u8; 65].tap_mut(...) buffer construction.
- Introduced module-private INTERNAL / LEAF u8 constants, used as both the hashing domain prefix and the encode/decode wire tag. Single source of truth for the variant tag values.
- Dropped tap::Tap import from node.rs (tap still used in key.rs, so the workspace dep stays).
- core/smt/Cargo.toml: swapped direct blake3 dep for vprogs-core-hashing.

External consumers (Cargo.toml + import updates):
- storage/rocksdb-store
- zk/abi
- scheduling/scheduler (dev-deps)
- core/smt/tests/e2e.rs (split imports: Blake3 / Hasher from new crate, EMPTY_HASH from vprogs_core_smt)

Some &value → value clippy fixes fell out of the impl AsRef<[u8]> signature change.

Root Cargo.toml:
- Added vprogs-core-hashing = { path = "core/hashing" } to workspace.dependencies. The core/* glob in members / default-members already covers the new crate.


core/hashing/src/blake3.rs
```diff
@@ -0,0 +1,22 @@
...
+    fn hash(data: impl AsRef<[u8]>) -> [u8; 32] {
+        *blake3::hash(data.as_ref()).as_bytes()
+    }
+
+    fn hash_parts_with_domain<const N: usize>(
+        domain: &[u8; N],
+        parts: impl IntoIterator<Item = impl AsRef<[u8]>>,
+    ) -> [u8; 32] {
+        let mut hasher = blake3::Hasher::new();
+        hasher.update(domain);
```
biryukovmaxim: shouldnt we use keyed hasher for domain separation

hmoog: Changed it

biryukovmaxim (approved):

PR #52, opened 2026-05-18
Feat: Wire Groth16 settlement covenant + backend tag

biryukovmaxim:
- Replace `RedeemPins` flat struct with `RedeemPins` enum (Succinct/Groth16 variants over `CommonPins`); each variant carries only the fields its verify branch consumes. Adds `verify_risc0_groth16` helper with inline `compute_receipt_claim` and `split_at_mid` (mirrors examples/zk-covenant-common). Removes the `panic!("groth16 settlement not wired yet")` at the old match arm.
- Add `covenant/build.rs` that precomputes the Groth16 verifier identity (compressed VK, control-root halves, bn254 control id, three risc0 tag digests) and emits them as `const` literals included by `script.rs`. The runtime covenant crate stays free of arkworks / risc0-groth16 deps.
- `SettlementWitness` enum replaces direct `SuccinctWitness` field on `SettlementInput`. `Settlement::build` branches on pins/witness variant and picks the right sig_script builder; mismatched variants panic.
- `Backend::new(tx_elf, batch_elf, zk_tag)` is now tag-aware. `prove_batch` selects `ProverOpts::groth16()` vs `succinct()` from the backend's tag; transaction proving stays succinct (inner receipts compose as assumptions). Adds `Backend::zk_tag()` getter + `OwnedGroth16Witness` for receipt-to-witness extraction.
- Unit tests parameterized over both proof systems. Adds cross-tests that re-derive each baked Groth16 constant from live risc0-groth16 / risc0-zkvm at test time and assert byte-for-byte equality, so risc0 ABI shifts trip the test before the on-chain Groth16 precompile silently breaks. Adds a `batch_proof_groth16_is_directly_settleable_single_batch` e2e mirror that exercises the full Groth16 path through Backend -> receipt -> settlement.


zk/backend/risc0/api/src/backend.rs
```diff
@@ -27,14 +28,23 @@ pub struct Backend {
     batch_elf: Vec<u8>,
     /// Batch processor guest image ID - the covenant script pins against this id.
     batch_image_id: [u8; 32],
+    /// Proof system the batch prover terminates in. Selects `ProverOpts::succinct()` vs
+    /// `ProverOpts::groth16()` and is exposed to host orchestration so the matching
+    /// covenant `RedeemPins` variant can be built. Transaction proving always uses succinct
+    /// regardless — inner tx receipts are folded into the batch as succinct assumptions; only
+    /// the outer batch receipt's kind varies.
+    zk_tag: ZkTag,
```
hmoog: Why not call this ProofType instead of ZkTag and the property settlement_proof_type or sth? 

Then we need less detailed comments and it becomes a bit more self-explanatory or is zktag some domain specific term that we want to mimic?

I would most probably also define the proof type in the api crate - e.g. proof_type.rs and reference it from the covenant instead of the other way round.

hmoog (changes requested):
LGTM, apart from the one question I added.

zk/backend/risc0/api/src/backend.rs
biryukovmaxim: I just reused what i already have, but proof type makes more sense

biryukovmaxim: fixed https://github.com/kaspanet/vprogs/pull/52/changes/633e1d329df6e2cfe026b2098b25dc72d0efc014

hmoog (approved):

PR #54, opened 2026-05-20
Feat: L1 bridge - settlement tracking

hmoog:
## Summary

  - Add `SettlementInfo` + `L1TransactionCovenantExt::settlement_info(covenant_id, containing_block)` to decode covenant settlements from accepted L1 transactions.
  - Track the most-recent settlement of a configured covenant on `ChainBlockMetadata::last_settlement`, carried forward across blocks with no new settlement.
  - New `L1BridgeConfig::covenant_id` (+ `--l1-bridge-covenant-id` CLI flag) opts the bridge into tracking; default `None` keeps existing behaviour.

### DIFF
--- a/l1/bridge/src/worker.rs
+++ b/l1/bridge/src/worker.rs
@@ -59,8 +59,7 @@ pub(crate) struct BridgeWorker {
     lane_key: Option<Hash>,
     /// Blue-score window within which a lane stays active without new transactions.
     finality_depth: u64,
-    /// Covenant id whose last settlement gets tracked by `ChainBlockMetadata::last_settlement` or
-    /// `None` to disable covenant tracking.
+    /// Covenant id tracked by [`ChainBlockMetadata::last_settlement`], or `None` to disable.
     covenant_id: Option<Hash>,
 }
 
--- a/l1/bridge/src/worker.rs
+++ b/l1/bridge/src/worker.rs
@@ -19,7 +19,7 @@ use kaspa_seq_commit::{
 use kaspa_wrpc_client::prelude::*;
 use tokio::sync::Notify;
 use vprogs_core_types::Checkpoint;
-use vprogs_l1_types::{ChainBlockMetadata, Hash, L1Transaction};
+use vprogs_l1_types::{ChainBlockMetadata, Hash, L1Transaction, L1TransactionCovenantExt};
 use workflow_core::channel::{Channel, MultiplexerChannel};
 
 use crate::{
@@ -59,6 +59,9 @@ pub(crate) struct BridgeWorker {
     lane_key: Option<Hash>,
     /// Blue-score window within which a lane stays active without new transactions.
     finality_depth: u64,
+    /// Covenant id whose last settlement gets tracked by `ChainBlockMetadata::last_settlement` or
+    /// `None` to disable covenant tracking.
+    covenant_id: Option<Hash>,
 }
 
 impl BridgeWorker {
@@ -138,6 +141,7 @@ impl BridgeWorker {
             subnetwork_filter: config.subnetwork_id,
             lane_key,
             finality_depth: config.finality_depth,
+            covenant_id: config.covenant_id,
         }
         .run()
         .await;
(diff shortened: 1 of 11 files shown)


l1/bridge/src/worker.rs
```diff
@@ -59,6 +59,8 @@ pub(crate) struct BridgeWorker {
     lane_key: Option<Hash>,
     /// Blue-score window within which a lane stays active without new transactions.
     finality_depth: u64,
+    /// Covenant id tracked by [`ChainBlockMetadata::last_settlement`], or `None` to disable.
+    covenant_id: Option<Hash>,
```
biryukovmaxim, line 63: In what cases covenant id is not required?

l1/types/src/l1_transaction_covenant_ext.rs
```diff
@@ -0,0 +1,267 @@
...
+/// Returns the payloads of the last four data pushes in `script`, or `None` if `script` contains
+/// fewer than four pushes or any byte is a non-push opcode.
+///
+/// Kaspa data-push opcodes (mirrors `kaspa_txscript::opcodes::codes`):
+/// - `0x00`: empty push.
+/// - `0x01..=0x4b`: literal push, opcode byte is the payload length (`OpData1..OpData75`).
+/// - `0x4c`: `OpPushData1`, 1-byte length prefix.
+/// - `0x4d`: `OpPushData2`, 2-byte little-endian length prefix.
+/// - `0x4e`: `OpPushData4`, 4-byte little-endian length prefix.
+fn sig_script_tail_4(script: &[u8]) -> Option<[&[u8]; 4]> {
```
biryukovmaxim: Can't we use Kaspa-txscript as dependency for parsing?

I didn't get why do we need to parse script into [&[u8];4] first and then convert it into [u8;32]. Can't we parse it directly into expected structure

l1/types/src/settlement_info.rs
```diff
@@ -0,0 +1,21 @@
+use borsh::{BorshDeserialize, BorshSerialize};
+
+use crate::{Hash, TransactionId};
+
+/// Decoded view of a settlement transaction observed on L1.
```
biryukovmaxim, line 6: Usually "view" used for zerocopy representations. Which also may be used here

node/cli/src/l1_bridge_params.rs
```diff
@@ -37,6 +37,10 @@ pub struct L1BridgeParams {
     /// Blue-score window within which a lane stays active without new transactions.
     #[arg(long = "l1-bridge-finality-depth", default_value_t = L1BridgeConfig::default().finality_depth)]
     pub finality_depth: u64,
+    /// Covenant id (64-char hex) whose last settlement gets tracked by each chain block's
+    /// metadata. Omit to disable covenant tracking.
+    #[arg(long = "l1-bridge-covenant-id")]
+    pub covenant_id: Option<Hash>,
```
biryukovmaxim, line 43: Shouldn't we require to pass the Arg?

biryukovmaxim (changes requested):

l1/bridge/src/worker.rs
hmoog: The bridge is a self-contained software crate and if you want to use it to build sth else - e.g. an indexer that doesn't want to settle anything or sth vprogs unrelated then you can also do that.

But in our case we will have it be a mandatory field (but that will become a requirement through the node cli-param mapping).

node/cli/src/l1_bridge_params.rs
hmoog: Yeah, once we wire up the zk-vm in the node this will become a mandatory field but I didnt want to break or touch existing tests for now - we can change this to a mandatory field once we wire this up.

l1/types/src/l1_transaction_covenant_ext.rs
hmoog: Yes, we can - it allocates a boxed opcode per push (~8 per succinct sig_script) which is slightly less efficient but considering that we only see a small number of settlements compared to the rest of the logic, this is anyway negligible and most probably better / less error prone.

I will change it.

biryukovmaxim: i think if we ever consider it a bottleneck - its better to update upstream to make it more efficient

biryukovmaxim (approved):

PR #57, opened 2026-05-22
Commit to lane

biryukovmaxim:


zk/abi/src/batch_processor/input/inputs.rs
```diff
@@ -19,13 +18,16 @@ use crate::{
...
+/// The lane this bundle settles is pinned at compile time (the batch-processor binary bakes its
+/// `LANE_ID` const), so no `lane_key` host input appears here: the value the guest uses is derived
+/// from the baked const, and the covenant SPK pins the same subnetwork id in its redeem-script
+/// prefix.
 pub struct Inputs<'a> {
     /// Transaction processor guest image ID used to verify each inner tx journal.
     pub image_id: &'a [u8; 32],
     /// Covenant id this bundle settles into.
     pub covenant_id: &'a [u8; 32],
-    /// Lane key for this bundle (one lane per bundle).
```
hmoog, line 27: I don't think we should hard-code this. By making it a public input to the circuit, we can reuse the same circuits for different guests which is ultimately our goal to standardize processing around a few whitelisted and interoperable guest circuits eventually once we advance to composability.

Having each guest result in completely different hard-wired circuitry that has to be reviewed and vetted again and again is extremely undesirable.

For now we wont get around the fact that people will have to audit the actual guest logic but everything around it should eventually turn into "generic and standardized circuitry" that doesn't deviate from guest to guest.

hmoog (changes requested):

zk/abi/src/batch_processor/input/inputs.rs
biryukovmaxim: updated

zk/abi/src/batch_processor/journal/state_transition.rs
```diff
@@ -24,8 +24,15 @@ pub struct StateTransition {
...
     /// output for permission-tree withdrawals.
     pub permission_spk_hash: [u8; 32],
+    /// Kaspa SubnetworkId this settlement binds to (the lane the batch processor was built for).
+    /// The covenant SPK pins the same 20 bytes in its redeem-script prefix, so any settlement
+    /// proof that names a different lane fails the SHA-256 journal-preimage check.
+    pub subnetwork_id: [u8; 20],
 }
 
+/// Serialized length of [`StateTransition`] in bytes.
+pub const JOURNAL_SIZE: usize = 32 * 8 + 20;
```
hmoog: We should use `core::mem::size_of::<StateTransition>()` instead of a hand-rolled constant.

It's safe because StateTransition derives Unaligned, which is a zerocopy invariant that the type has alignment 1 and therefore no padding bytes between fields. Combined with #[repr(C)] + KnownLayout, that means the in-memory size equals the sum of field byte widths equals the wire-encoded length. 

size_of is const fn, so this works wherever the literal works.

zk/abi/src/batch_processor/input/inputs.rs
```diff
@@ -19,13 +18,21 @@ use crate::{
...
-    pub lane_key: &'a Hash,
+    /// Kaspa SubnetworkId of the lane this bundle settles. Committed to the journal and re-derived
+    /// into `lane_key`; the covenant SPK pins the same 20 bytes in its redeem-script prefix.
+    pub subnetwork_id: &'a [u8; 20],
```
hmoog: I know we said that it might make more sense to pass in the subnetwork_id instead of the lane_key but it kind of sucks that we have to use the array representation here instead of the SubnetworkId type and lose type honesty.

I understand that this is because kaspas upstream crates that hold the SubnetworkId type are not no-std compatible but maybe this is a hint that we should actually continue to use the lane_key instead.

Also because lanes are the actual concepts for the L2 and while the L1 currently transparently maps the subnetwork_id to the lane, this might change in the future if we introduce sub-lanes or other identifiers that might determine the lane_key.

It feels like the lane_key while currently strictly less powerful for derivation (because you can go from subnetwork_id to lane_key - but not the other way round) is maybe still the better concept for forward compatibility (and it also saves us another hash in the guest as we only actually need the lane_key for commitments and stuff).

What do you think?

zk/abi/src/batch_processor/lane.rs
```diff
@@ -0,0 +1,55 @@
+//! Lane-identity helpers.
```
hmoog: This entire file feels like it doesn't belong into the ABI as its just a helper to construct subnetwork_ids (even with the wrong return type due to the no-std limitation I mentioned earlier).

It essentially just deals with upstream L1 types that are not part of the zk-abi interface (how guests and the zk system communicates and none of the guests needs it).

It feels like it might rather live as an extension trait in the test-suite or in the l1 domain but `consensus-core/src/subnets.rs` already has a `pub const fn from_namespace(namespace: [u8; SUBNETWORK_NAMESPACE_LEN]) -> SubnetworkId {` function which could be invoked with `SubnetworkId::from_namespace(4444u32.to_be_bytes())`, without having to introduce this additional helper.

I see it has an additional assert but this assert also feels like it should live further up in the layered approach (e.g. at the place where we parse node inputs eventually).

zk/abi/src/batch_processor/verifier.rs
```diff
@@ -23,13 +23,21 @@ use crate::{
 };
 
 /// Verifies a bundle and accumulates its post-state.
+///
```
hmoog: This spells out field related logic and how its used in the type comment.

I personally prefer to not re-explain code in a type associated comment because this can change as part of refactors (e.g. the upcoming settlement / batch processing logic split) which would leave these kind of comment out of sync.

Imho it makes more sense to explain this relationship in the covenant / StateTransition (journal) than the verifier but I guess its a bit of a style preference thing so feel free to object and we can resolve this.

zk/backend/risc0/batch-processor/src/main.rs
```diff
@@ -10,7 +10,8 @@ risc0_zkvm::guest::entry!(main);
 /// Entrypoint for the batch processor.
 fn main() {
     // Read the bundle inputs from the host and build the verifier with the default
-    // permission-tree accumulator for exits.
+    // permission-tree accumulator for exits. The lane the bundle settles arrives as a public
```
hmoog: This comment change feels out of place that justifies unrelated changes and I am not sure it makes sense to specifically name the subnetwork_id here

zk/backend/risc0/test-suite/tests/settlement_e2e.rs
```diff
@@ -31,6 +34,10 @@ use vprogs_zk_backend_risc0_test_suite::{
     compute_section_lane_tip, dev_mode_enabled, transaction_processor_elf,
     transaction_processor_with_exits_elf,
 };
+
+/// Subnetwork id all e2e settlement fixtures bind to; must match the `LANE_ID` baked into the batch
+/// processor guest binary.
+const TEST_SUBNETWORK_ID: [u8; 20] = subnetwork_id_from_lane_id(4444);
```
hmoog: the TEST_SUBNETWORK_ID is defined multiple times in different tests - maybe it would make more sense to expose it once as part of test-suite/src/consts.src if it's even necessary in relation to some of the previous comments.

hmoog (commented):
Some further comments - feel free to object / comment.

hmoog (changes requested):

zk/abi/src/batch_processor/input/inputs.rs
biryukovmaxim: Im fine to revert it back. also i see no issue to keep subnetwork id since there is no clear picture if and when it will be changed in future

zk/batch-prover/src/config.rs
```diff
@@ -13,8 +14,8 @@ pub struct BatchProverConfig {
     /// amortize settlement cost; smaller bundles bound the worst-case wasted compute on a
     /// reorg-induced retry.
     pub bundle_size: NonZeroUsize,
-    /// Our lane key. Bundle-wide (one lane per prover instance).
-    pub lane_key: Hash,
+    /// The lane this prover settles; bundle-wide (one lane per prover instance).
```
hmoog: Shouldn't this also become a lane_key again to be consistent with the needs of the proving circuit?

zk/abi/src/batch_processor/verifier.rs
```diff
@@ -30,6 +30,9 @@ where
 {
     /// Decoded bundle inputs.
     inputs: Inputs<'a>,
+    /// Lane key for this bundle: the SMT key for this lane in L1's `lanes_root` and the lane
+    /// domain in `lane_tip_next`.
+    lane_key: Hash,
```
hmoog: We can access the lane_key using self.inputs.lane_key and don't need to copy it into the verifier.

hmoog (changes requested):
Just two last things - other than that it looks good.

hmoog (approved):
LGTM

PR #58, opened 2026-05-29
Bump rusty-kaspa and wrap lanes_root into the post-hardening activity…

biryukovmaxim:
… root

Update the rusty-kaspa dependency to the rebased covpp-reset2-rpc-lane-proof HEAD, which reshapes the seq-commit state root: the lanes_root is now wrapped into an activity_root via H_activity_root(inactivity_shortcut, lanes_root) before feeding seq_state_root.

- Carry inactivity_shortcut through LaneProof (decode/encode from the RPC GetSeqCommitLaneProofResponse) and apply activity_root_hash in new_seq_commit.
- Drop the removed seq_commit_timestamp identity helper; hash prev_timestamp directly in the verifier and the test-suite lane-tip helper.
- Track the Params rename in the L1 e2e/covenant tests: covenants_activation becomes toccata_activation + zk_hardening_activation (both always-on), and block_mass_limits is set via prior_block_mass_limits.
- Refresh the batch- and transaction-processor guest ELFs.


hmoog (approved):
Looks pretty straight forward - imho we can just merge as is.

Btw. what's up with the seq_commit_timestamp function? That feels like a completely pointless function? Is it just there to name tag things in upstream or sth?

biryukovmaxim:
> Looks pretty straight forward - imho we can just merge as is.
> 
> Btw. what's up with the seq_commit_timestamp function? That feels like a completely pointless function? Is it just there to name tag things in upstream or sth?

maybe initiall

> Looks pretty straight forward - imho we can just merge as is.
> 
> Btw. what's up with the seq_commit_timestamp function? That feels like a completely pointless function? Is it just there to name tag things in upstream or sth?

probably initially i used it to point that it must accept parent timestamp. but it was removed during cleanup in rk

PR #72, opened 2026-06-04
Feat: Batch aggregator split

hmoog:
## Summary

Splits the monolithic batch verifier into **two composable risc0 guests** — a per-batch **batch-processor** and a bundle **batch-aggregator** — so each batch is proven independently and the aggregator composes them into a single settlement proof. Introduces a new intermediate identity (`batch_image_id`) and binds it on-chain, reorganizes the L2→L1 exit types into a shared `withdrawal` module, and consolidates supporting plumbing.

Base: `feat/zk-convenants`. ~65 files, +3.4k/−1.1k (incl. regenerated guest ELFs and lockfiles).

## Why

Previously a single circuit verified an entire bundle in one pass. Splitting it lets each batch be proven as soon as its txs are ready and lets the aggregator compose a variable number of per-batch receipts via `env::verify` — better parallelism and a cleaner trust boundary, at the cost of one extra composition layer that must be cryptographically bound.

### Circuit split
- **New `batch-aggregator` guest** (`zk/abi/src/batch_aggregator/`, `zk/backend/risc0/batch-aggregator/`): `env::verify`s each per-batch receipt against `batch_image_id`, chains the `BatchTransition`s (each batch's `prev_*` must equal the previous batch's `new_*`), streams the bundle's exits into a permission tree, derives the final `new_seq_commit`, and commits a `StateTransition` settlement journal.
- **`batch-processor` now emits a per-batch `BatchTransition` journal** (`zk/abi/src/batch_processor/`) instead of a bundle-wide journal; it `env::verify`s its tx receipts against `tx_image_id` and commits that id.
- The old monolithic per-bundle assertions are preserved, redistributed across the two guests: per-tx checks (merge_idx monotonicity, version-shape, context_hash, resource hash/id) stay in the batch processor; the new inter-batch chaining + per-bundle invariant checks (`lane_key`/`covenant_id`/`tx_image_id` constant) live in the aggregator.

### On-chain binding
- The covenant redeem script now pins **three** image ids — `program_id` (= aggregator), `batch_image_id`, and `tx_image_id` — and cats `batch_image_id` into the journal preimage (288 → **320 bytes**). This closes the new trust boundary: without the `batch_image_id` pin, a host could compose against a malicious batch processor undetected.

### Supporting changes
- **`withdrawal` module** (`zk/abi/src/withdrawal/`): the exit vocabulary (`ExitSink`, `Exits`/`ExitsIter`, `ExitAccumulator`/`NoExits`, `StandardSpk`, `ScriptBytes`) moved out of the transaction-processor journal into a shared, guest-independent module.
- **Single `ErrorCode` enum** (`zk/abi/src/error_code.rs`): the per-module guest error codes were merged into one crate-level enum so discriminant uniqueness (committed in journals) is compiler-enforced. Values unchanged (`VersionIncompatible=1`, `InvalidExitSpkTag=2`, `InvalidExitSpk=3`).
- **Generalized `Journals` iterator** (`zk/abi/src/journals.rs`): one length-prefixed-slice iterator shared by both the batch-processor (tx journals) and aggregator (batch journals).
- **Zero-alloc borrowed SMT-proof decode**: the aggregator decodes the lane SMT proof into a borrowed `kaspa_smt::SmtProof` at input-decode time (no `Vec` allocation in the guest) instead of `OwnedSmtProof::from_bytes`.
- **`Backend` refactor** (`zk/backend/risc0/api/`): the three `(elf, image_id)` pairs are consolidated into an `ElfBinary` type; trivial getters dropped in favor of public read-only fields.
- **Named journal encode inputs**: `StateTransition::encode` / `BatchTransition::encode` now take borrowed, named `StateTransitionArgs` / `BatchTransitionArgs` structs (co-located with each journal type, with per-field `See [...]` doc-links) instead of positional `&[u8; 32]`/`&Hash` tuples, so a transposed field surfaces in review rather than as a silent wire corruption. The free `JOURNAL_SIZE` const is now the associated `StateTransition::WIRE_SIZE`. `BatchTransition`'s fields were also reordered to group `prev_*` then `new_*` (matching `StateTransition`), which changes the batch journal wire format — safe while WIP since it has no pinned consumers and encode, the `ref_from_bytes` decode, and the field-name-based verifier reads all stay in lockstep.
- **Batch-prover commit ordering fix** (`zk/batch-prover/src/worker.rs`): the prover worker now awaits each batch's commit (`wait_committed`) before pulling the next, restoring a barrier dropped in the per-batch worker rewrite. Without it a batch could prove its `prev_state` against an SMT version the predecessor hadn't committed yet — `Tree::node` silently falls back to an older root, authenticating a stale `prev_state` that the aggregator's `prev_state == prev.new_state` check then rejects after both proofs were already paid for.

### Dependency
- **kaspa fork bump `1.2.1-toc.3` → `1.3.0-toc.5`** (`covpp-reset2-rpc-lane-proof` @ `aa6ff3da`), which exposes the borrowed `SmtProof` (marker removed) and public terminal-tag consts. Fallout: `Params::zk_hardening_activation` was removed in 1.3 (folded into `toccata`) — dropped from the test fork-activation setup; `toccata_activation = always()` retained.

## Cryptographic binding

The split's safety hinges on every image id used for an `env::verify` being committed-into-a-journal and pinned on-chain. The chain holds end-to-end:

- `tx_image_id` — verified in the batch processor, committed in `BatchTransition`, asserted-shared across the bundle by the aggregator, committed in `StateTransition`, pinned by the covenant.
- `batch_image_id` — verified in the aggregator (same value committed into `StateTransition`), pinned by the covenant.
- `aggregator_image_id` — the circuit `OpZkPrecompile` verifies (covenant `program_id`).

The covenant's 320-byte SHA-256 journal preimage matches `StateTransition::encode` field-for-field, so a proof committing any other image id fails the on-chain hash check. No "verified-but-not-pinned" path exists.

## Testing

- New `zk/abi/tests/batch_aggregator.rs` exercises the aggregator's chain conditions, per-bundle invariants, exit ordering, and `lane_expired` handling against hand-built `BatchTransition` journals (no proving).
- Settlement / covenant e2e suites updated for the new image-id wiring and the 320-byte journal.

## Known follow-ups (WIP)

- **Aggregation is not yet wired into the production proving pipeline** — `prove_aggregator` and the `get_seq_commit_lane_proof` fetch that feeds `new_seq_commit` are currently exercised only via the test helper. The lane-proof fetch must be re-introduced when the aggregation orchestrator lands.
- **Worker `Command::Rollback` is a no-op stub** (`// TODO`): a rolled-back batch can still be proved/published. Host-side only (the covenant still gates on-chain correctness).


zk/abi/src/batch_aggregator/journal/state_transition.rs
```diff
@@ -41,7 +43,7 @@ impl StateTransition {
         (prev_state, prev_lane_tip): (&[u8; 32], &Hash),
         (new_state, new_lane_tip, new_seq_commit): (&[u8; 32], &Hash, &Hash),
         covenant_id: &[u8; 32],
-        tx_image_id: &[u8; 32],
+        (tx_image_id, batch_image_id): (&[u8; 32], &[u8; 32]),
```
biryukovmaxim: shouldnt we already introduce structure? for input args? all types almost the same, too easy to misuse from caller side

biryukovmaxim (changes requested):

biryukovmaxim:
**🟠 Worker regression: batch proving no longer waits for the predecessor's commit → stale `prev_state`, wasted proofs, aggregator panic**

The rewrite of `zk/batch-prover/src/worker.rs` dropped the commit-wait barrier that the old bundle worker had. The old `process_bundle` ended with:

```rust
// Wait for the bundle's final commit before pulling the next bundle.
for batch in bundle.batches() {
    batch.wait_committed().await;
}
```

The new `process_batch` (`worker.rs:66-95`) has no equivalent — it proves at `prev_version = checkpoint().index() - 1` (`worker.rs:74-75`), publishes, and immediately pulls the next batch.

**Why this is a correctness bug, not just a missing nicety**

The prover and the commit path run on independent threads, both fed at schedule time, with nothing synchronizing them:

- `scheduler.schedule()` pushes the batch to the prover inbox via `on_batch_scheduled` (`scheduler.rs:79` → `proving_pipeline.rs:49-52` → `prover.rs:40-42`) **before** the batch is committed.
- Commit runs separately on the lifecycle worker (`batch_lifecycle_worker.rs:48-52`); batch N writes SMT **version N** in `commit()` (`scheduled_batch.rs:349-374`), opening the `committed` latch only in `commit_done` (`:383`).
- `Tree::node` resolves *"latest version ≤ max_version"* (`core/smt/src/tree.rs:24-26`). So if batch N-1's commit hasn't landed, `store.prove(keys, N-1)` silently authenticates against **root(N-2)** and stale leaf values — no error.
- The batch-processor guest commits that as `prev_state`. The aggregator then asserts `this.prev_state == prev.new_state` (`zk/abi/src/batch_aggregator/verifier.rs:81`) → root(N-2) ≠ root(N-1) → **panic, after both batches were already proved (paid for).**

This is masked today only because proving (seconds) outlasts commit (a fast write). The dependency is incidental timing, not an enforced happens-before — a slower disk, a larger commit, or anything that speeds proving / slows commit flips it.

**Reproduction** (deterministic, isolates the stale read the worker now performs):

```rust
// core/smt/tests/prove_ahead_of_commit.rs
#[test]
fn prove_at_uncommitted_version_serves_stale_state() {
    let dir = TempDir::new().expect("temp dir");
    let store: RocksDbStore = RocksDbStore::open(dir.path());
    let r = ResourceId::from([1u8; 32]);
    let (v1, v2) = ([0x11u8; 32], [0x22u8; 32]);

    // Batch 1 committed at version 1.
    let root1 = commit(&store, 1, &[(r, v1)]);

    // Batch 2's commit (version 2) hasn't landed; the worker proves at prev_version = 2.
    // root(2) silently falls back to root(1) — no signal version 2 is missing.
    assert_eq!(store.root(2), root1);

    // prove(_, 2) authenticates the STALE v1 leaf.
    let proof = Proof::decode(&store.prove(&[r], 2).expect("prove")).expect("decode");
    assert_eq!(proof.leaves[0].value_hash, Sha256::hash(v1));

    // After batch 2 commits, the true root(2) differs — the worker proved against a stale prev_state.
    assert_ne!(commit(&store, 2, &[(r, v2)]), root1);
}
```

**Fix:** await the predecessor batch's `committed` latch before `store.prove` in `process_batch` (re-instating the old `wait_committed()` guarantee), so the worker never proves at a version that isn't yet committed.

zk/abi/src/batch_aggregator/journal/state_transition.rs
hmoog: I added *Args structs for the collection of parameters which is essentially a mirror of the owned pod style struct (so we can hand in references to the existing data without having to copy them).

biryukovmaxim (approved):

Issue #67, opened 2026-06-01
Reorg-aware proof orchestrator


hmoog:
## Summary

The three-level proving architecture (tx → batch → aggregator) delivers its reorg-resilience and fast-forward benefits only if a driver above the prover knows which batches survive a reorg and which need re-proving. Build the component that monitors the L1 tip, detects reorgs and external settlements, and orchestrates proof invalidation plus re-proving.

## Motivation

Without this driver, the split architecture's reorg benefit is theoretical: the prover keeps generating proofs but doesn't know when to skip vs redo. With the driver in place:

- Reorg at the tip → invalidate batches at or after the reorg point, keep the rest.
- External settlement (another prover settles ahead of us) → match our cache against their tip, prove only the suffix.
- Restart → reload cache from the proof store and resume.

## Design

```
+----------------------+
| L1 watcher / sync    |  emits chain events: new blocks, reorgs, settlements
+-----------+----------+
            |
            v
+----------------------+
| Reorg orchestrator   |  decides which batches survive, which to re-prove
+-----------+----------+
            |
            v
+----------------------+
| Prover driver        |  uses proof store (issue #7) to skip / generate
+----------------------+
```

Inputs:
- L1 chain state: current tip, recent ancestry.
- Local proof store (issue #66).
- Pending batches awaiting proof.

Outputs:
- Invalidation calls to the proof store for batches no longer on the canonical chain.
- New batch-prover jobs for batches now on the canonical chain but not yet proven.
- (Eventually) Aggregator jobs to re-aggregate over the surviving + new batch proofs.

## Scope

- New crate / module: `zk/orchestrator/...` (or wherever the prover driver lives).
- Integrate with the L1 watcher used for settlement tracking.
- Hook into the proof store (issue #66) for cache lookups and invalidation.
- Tests: simulate a reorg at depth K, verify the right batches get invalidated and re-prover jobs are queued; simulate an external settlement, verify the suffix-only proving path triggers.

## Acceptance

- Simulated reorg at depth K invalidates only batches at or after the reorg point.
- Unaffected batches remain in the proof store and aren't re-proven.
- After re-proving the affected suffix, the aggregator (issue #64) produces a valid settlement.
- Simulated external settlement triggers the suffix-only proving path.

## Out of scope

- Multi-prover coordination, proof gossip, or an on-chain proof registry (future concern; can be noted in body for future expansion).
- Batch-aggregator circuit itself (issue #64, prerequisite).
[...]

PR #73, opened 2026-06-08
Testnet 10 flow

biryukovmaxim:
## Summary

Adds two ways to exercise the full L2 flow end-to-end — a live POC daemon against a remote
testnet-10 fork node, and a seeded deterministic simulation against an in-process kaspa DAG — and
factors out the shared L1 transaction-building they both need. Each follows the selected chain,
schedules lane txs through the zk `Vm`, executes them, and (optionally) proves batches and settles
them on chain through a production covenant. There is no behavior change to the existing
settlement/proving path; the new crates sit alongside it and the edits to existing crates are
decoupling seams (plus a constructor-signature update threaded through the tests).

## `examples/tn10-flow` — live POC daemon

One env-driven process that bootstraps (or reuses) a covenant, periodically issues lane activity,
and follows the chain via the existing `L1Bridge`. In execution-only mode (default) it runs a
framework `Node` with `ProvingPipeline::None` that tracks the decoded state counter, reorgs, and
settlements, with the worker's and Vm's `trace` logs as the per-block observability. With
`TN10_SETTLE=1` it drives the real batch prover and forwards each scheduled batch to the settler,
which proves every bundle and settles it with a production `Settlement::build` (real proofs need a
GPU — the `cuda` feature, without `RISC0_DEV_MODE`). State (lane id, covenant id, bootstrap txid)
persists to a JSON file so the daemon resumes across restarts.

## `sim` — seeded deterministic L2 simulation

Runs the full L2 stack (scheduler → zk `Vm` → optional proving → settlement) against a simulated
kaspa DAG built from real `TestConsensus` nodes, asserting invariants every block; a fixed seed
reproduces any failure. It reuses the rusty-kaspa/simpa fork's discrete-event engine and `Miner`
mechanics as libraries (the fork is consumed as a pinned git dependency, never edited). The driver
follows the selected chain, executes lane txs, rolls back on reorg, and runs the covenant lifecycle
(bootstrap → chained settlements), with the counter invariant and settlement acceptance checked on
every block. The dev-mode tests cover execution, determinism, reorgs, and dev settlements on CPU; a
GPU-gated test chains real-proof, real-on-chain settlements — the same shape as `settlement_l1_e2e`
but driven by simpa instead of `L1Node` + RPC.

## `l1/wallet` — shared L1 tx building

Pure, RPC-free transaction constructors (activity, covenant bootstrap, and settlement funding, with
the fee and mass logic) plus a thin `Wallet<C: RpcApi>` that fetches spendable UTXOs and submits.
Lifted out of `node/test-utils`'s `L1Node` (mostly deletions there) so the example and the sim share
one implementation; an in-process caller that already holds its UTXO set uses the pure builders with
no RPC.

## Enabling seams in existing crates

The batch prover gains a `LaneProofSource` trait so it is no longer hardwired to `GrpcClient` — the
example reads lane proofs over wRPC, the sim reads the in-process consensus directly. The framework
worker forwards each scheduled or rolled-back batch to an optional `BatchEvent` sink, which is how
the settler receives the bundles it settles. The bridge gains `seed_from_sink` so a fresh lane can
follow from "now" rather than genesis. And the covenant's settlement input now sizes its compute
budget by actually running the script engine (`covenant_compute_budget`) instead of a hand-tuned
constant.

## Suggested review order

Intent → shared building blocks → seams → the two deliverables → tests: `main.rs` doc → `l1/wallet`
(lib + build) → `LaneProofSource` → `settlement.rs` budget → `settler.rs` → `sim/src/driver.rs` →
`sim/tests/l2_flow.rs`.


examples/tn10-flow/src/settler.rs
biryukovmaxim, line 1: probably will be rewritten or migrated as part or on top of https://github.com/kaspanet/vprogs/issues/67

PR #80, opened 2026-06-16
Wire aggregation and settlement into the proving pipeline

biryukovmaxim:
## Summary

Carries the L2 pipeline through to a **settlement-level receipt**: the framework follows the
chain, executes, and proves transaction → batch → bundle, then hands each bundle to a
separate settlement worker that lands it on L1. Adds the missing pipeline stages — an
**aggregate prover** that forms bundles, a **settler** that submits them, and a fork-aware
**proof-receipt store** — plus a seeded **L2 simulation** crate and a `tn10-flow` reference
binary that exercise the whole flow (including real proofs under cuda) against a reorging DAG.

Base: `feat/zk-convenants` (which now contains the squash-merged #72 batch-aggregator split).
This branch was updated by merging that base in; the merge tree is identical to the branch's
pre-merge tip, so the diff below is purely the proving-pipeline work on top of #72.

`docs/proving-pipeline.md` is the control-flow map for everything here and doubles as the gap
list.

## Why

After the aggregator split (#72) the per-batch and bundle circuits exist, but nothing drives
them end-to-end in process or settles the result on chain. This wires the three provers into
one `ProvingPipeline`, gives bundles a place to land (`BundleOutcome` → settlement sink), and
adds the persistence and simulation needed to run and test the flow across reorgs.

Key idea: **the framework proves all the way to the settlement receipt but does not submit
settlements.** The provers live inside the pipeline; a separate worker drains the bundle
outcomes and lands settlements on L1, so submission policy stays out of the proving path.

### Proving pipeline (`zk/vm`)
- `ProvingPipeline` composes `TransactionProver` (1 receipt/tx), `BatchProver`
  (1 receipt/batch) and the new `AggregateProver` (1 receipt/bundle). The aggregate prover
  forms bundles and emits each one's `BundleOutcome` (a `SettlementArtifact`, or a no-op
  marker for empty/all-empty bundles) on an optional sink.
- Bundling and settlement moved *into* the pipeline; outcomes are published to an async
  queue rather than shipped over a one-shot channel.

### Aggregate prover (`zk/aggregate-prover`)
- New crate: one worker thread, one bundle receipt per scheduled bundle, with its own
  command/queue plumbing. Drops rolled-back batches from its queue on reorg.

### Settlement worker (`zk/backend/risc0/settler`)
- New crate that drains the `BundleOutcome` stream, runs `Settlement::build`, submits to L1,
  and confirms the continuation UTXO. The reference binary runs it as an async task; the
  simulation drives the same stream synchronously.
- `aggregate_batches` (the lane-proof fetch + aggregator-input encode + `prove_aggregator`
  step) is factored into the shared test-suite so the settler, the sim driver, and the e2e
  tests run one implementation.

### Proof-receipt store (`state/proof-receipt`)
- Single-CF store keyed by semantic proving coordinates, **fork-aware** via a `block_hash`
  field so a reorg can't serve a stale receipt. Keys share a `checkpoint_index` prefix
  (zerocopy `repr(C)`), so one prefix scan prunes every receipt for a checkpoint regardless
  of image id or block hash.

### Lane-proof fetch abstraction
- `LaneProofSource` trait generalizes the batch prover's `get_seq_commit_lane_proof` fetch
  over the production wRPC `GrpcClient` and the simulation's in-process source, so the prover
  no longer hard-codes a gRPC client.

### Reorg safety
- Covenant tracking is reorg-safe and exercised under a reorging DAG; in-flight settlement is
  drained before the sim halts; proof receipts are keyed by block hash. (Remaining gaps —
  batch-prover `Rollback` is a no-op, an already-submitted settlement is not recalled — are
  documented in `docs/proving-pipeline.md` §7–8.)

### Lifecycle
- Prover workers are joined on shutdown so the GPU prover tears down before the process's
  CUDA atexit, avoiding a "driver shutting down" abort.

### Simulation + reference binary
- **`sim`**: seeded, deterministic L2 simulation reusing `simpa` + `TestConsensus`, with an
  L2 miner, network model, and a driver that runs the full execute→prove→settle loop. Tests
  cover the L2 flow and real-proof settlement on chain.
- **`examples/tn10-flow`**: execution POC daemon with a real-proof settlement mode, hardened
  failure paths, and mass-based L1 fee pricing. L1 tx-building helpers factored out into the
  new **`l1/wallet`** crate.

## Testing

- `sim/tests/l2_flow.rs` and `sim/tests/smoke.rs` drive the seeded simulation end-to-end.
- Settlement e2e suites (`zk/backend/risc0/test-suite`) updated to the shared
  `aggregate_batches` path and the new image-id / lane-proof wiring; a real-proof e2e
  settlement test passes end-to-end (cuda).
- Per the project policy, dev-mode tests gate on `dev_mode_enabled()`; real proofs run only
  with dev mode off + cuda on.

## Known follow-ups (documented in `docs/proving-pipeline.md` §8)

- **Batch-prover `Command::Rollback` is a no-op stub** — a rolled-back batch can still be
  proved/published (host-side only; the covenant still gates on-chain correctness).
- **An already-submitted settlement is not recalled** on reorg, and the settler carries TODOs
  for fee-bumping a settlement that doesn't confirm in time. Single-miner / low-reorg only for now.
- **`node/vm` + `node/cli` remain the stubbed path**; the live stack is `zk/*` driven by
  `examples/tn10-flow`.


zk/backend/risc0/api/src/backend.rs
hmoog: We should most probably get rid of this impl block and move the logic to the newly introduced Trait impl method, so it mirrors the code for the other traits.

zk/aggregate-prover/src/scheduled_bundle.rs
```diff
@@ -0,0 +1,150 @@
+use std::sync::Arc;
```
hmoog, line 1: I would actually move this type somewhere else.

I see the scheduler as a generic standalone component that should only concern itself with delegating access to resources and not contain any proving related logic and types as that is something that is happening on an additional layer above. Maybe it could move into the aggregator-prover crate in the zk domain or sth?

What do you think?

hmoog (changes requested):
I flagged two things.

And we usually follow the convention that each type gets it own file (especially if the type is public).

Could you maybe break the files that contain multiple pub types down into their individual file_name.rs or do you dislike this convention?

The same is true for a lot of the covenant related stuff where we sometimes have a large number of types in a single file but I didn't flag this before because I kind of assumed that we can possibly clean that up later if necessary but maybe it makes sense to raise it now so we can agree on the convention going forward.

I think its just easier to parse if each type is isolated and doesn't get swamped in a lot of logic that is only indirectly related.

zk/aggregate-prover/src/scheduled_bundle.rs
biryukovmaxim: it actually was there couple commits before. but i see bundle the same compponent as scheduled tx or scheduled batch that's why i moved it here from aggregator. i probably can revert that commit

zk/backend/risc0/api/src/backend.rs
biryukovmaxim: tbh i dont see big difference. but i would like to be able to call method on structure without importing of the trait into scope. but that doesnt really matter a lot to me so i will fix according to ur guidance

biryukovmaxim:
> I flagged two things.
> 
> And we usually follow the convention that each type gets it own file (especially if the type is public).
> 
> Could you maybe break the files that contain multiple pub types down into their individual file_name.rs or do you dislike this convention?


it depends. I think it makes sense to split driver and covenant into many files. but splitting ‎state/proof-receipt/src/key.rs into many files feels to me wrong

hmoog (approved):
Looks good now

PR #82, opened 2026-06-16
Feat: Fork-aware canonical chain tracking

hmoog:
## Summary

Adds a **fork-aware canonical chain**: the source of truth for which competing batch is canonical, kept as a lock-free in-memory bit overlay and reconstructed from persisted batch metadata on startup.

Batches are identified by a monotonic, never-reused id. Competing L1 forks produce different batches at adjacent ids; the canonical chain records which ids form the live chain (the tip's ancestry). SMT reads, rollback, and pruning all consult it: an orphaned batch's writes are invisible to reads, and its state is retained rather than pruned, so a block that returns after a reorg reuses its existing id and finds its data still available.

The chain is derived from data each batch already commits rather than a dedicated column family, and this PR lands it end-to-end, including the read-side SMT filtering and the L1 bridge that drives the scheduler from the live chain.

## Design

Layered so the concurrency primitive, the canonical overlay, the domain manager, and the consumers are each isolated:

- **`core/atomics` : `AtomicRing<V>`** (new): a generic lock-free, growable, index-addressed rolling buffer with a prunable prefix, a truncatable suffix, per-entry copy-on-write (`fork` / `replace`), and amortized-doubling growth. `base`, `tip`, and the slot array are independent atomics, so the live window can advance on one thread while another prunes. Reads (`get` / `with`) are wait-free.
- **`storage/canonical-chain`** (new crate): the fork-aware chain itself.
  - `Bucket`: a fixed-size run of 4096 canonical bits (`1` = canonical), addressed by id.
  - `CanonicalChain` / `CanonicalChainSnapshot`: the overlay is a ring of `Arc<Bucket>` published through an `ArcSwap`. A hot zone (the tail and last-sealed bucket) absorbs shallow reorgs copy-free, deeper sealed buckets are copy-on-write, and finalized buckets are pruned off the head (and thereafter read as canonical). `snapshot()` hands readers an immutable, wait-free `is_canonical(id)` view.
  - `CanonicalChainManager`: the single-writer handle. An append-only log of batch metadata keyed by dense, never-reused ids, with a `block_hash -> id` reverse index. `append` assigns or revives an id (returning `AppendOutcome { id, is_new }`), `rollback` orphans the suffix, `finalize` drops finalized log entries and prunes the overlay. Its constructor reconstructs the canonical set by walking each block's `parent_id` back to the finalized floor.
- **`core/types`**: a `ChainSink` trait (`append` / `rollback` / `finalize` / `tip` / `metadata` / `id`), the seam the scheduler implements and the bridge drives. `BatchMetadata` gains `block_hash()` and `parent_id()`, the data the chain reconstructs from.
- **`storage`**: `Store::canonical_chain()` exposes the read oracle; `Store::canonical_chain_manager()` rebuilds the manager by replaying the persisted `BatchMetadata` (each batch's id is its stored index). No dedicated column family: the chain rides on data already committed per batch.

## Read-side filtering

`Tree` gains a `Snapshot` associated type. Each operation (`root` / `prove` / `update`) captures one snapshot up front and threads it through every `node` lookup, so a multi-node walk sees a single consistent canonical view instead of re-snapshotting per node. `RocksDbStore::node` skips non-canonical versions using that snapshot, so an orphaned write is invisible to reads while its predecessor stays live.

## Lifecycle wiring

| Event | In-memory (live, scheduler thread) | Durable |
|---|---|---|
| Schedule | `manager.append(metadata)` in `next_checkpoint`, assigning a never-reused id | - |
| Commit | - | batch metadata staged into the commit `WriteBatch` (the chain reconstructs from it) |
| Reorg | `manager.rollback(index)` in `cancel_and_rollback` | rollback pointers reverted in `Rollback::execute`, keyed by the captured snapshot |
| Finality | `manager.finalize(below)` in `pruning_worker` | state and finalized batch metadata pruned; orphaned vs canonical decided by the snapshot |
| Startup | `Store::canonical_chain_manager()` replays `BatchMetadata` | - |

The in-memory overlay updates the instant a batch is scheduled or rolled back, so the live view never lags a delayed disk flush. On restart the manager rebuilds the overlay from the persisted batch metadata.

## Consumers

- **`scheduling/scheduler`**: implements `ChainSink` and owns the `CanonicalChainManager`. Rollback and pruning consult the snapshot: an orphaned batch reclaims its own (now dead) write, a canonical batch reclaims the superseded predecessor.
- **`l1/bridge`**: follows the Kaspa selected-parent chain and drives the scheduler directly, scheduling each chain block, mapping reorgs onto the never-reused id space, and advancing finalization, with a reorg filter and automatic reconnection. `ChainBlockMetadata` (`l1/types`) implements `BatchMetadata`.
- **`node/framework`**: `Node` wires the bridge to the scheduler and exposes a lock-free query API.
- **`sim`**: a covenant and reorg simulation driver exercises the chain end-to-end against a real Kaspa node, including the real-proof settlement path.

## Concurrency

The overlay has a single writer (the manager) publishing immutable snapshots through an `ArcSwap`; readers are wait-free, and a captured snapshot never changes its answers even as the shared ring grows or head-prunes. Within `AtomicRing`, `base`, `tip`, and the slot array are separate atomics, and `grow` replaces only the slot array, so a concurrent prune is never clobbered.

## Follow-ups

A returning block reuses its id and its state is retained, but it is still re-executed. Skipping that re-execution, by pausing a block's execution and resuming it when the block returns, is a separate later step that builds on the retained state this PR provides.

## Testing

- `AtomicRing` unit tests: read-back, wrap-after-prune, grow-and-remap, suffix truncation, fork independence, and a non-`Clone` `with` read.
- Canonical-chain tests over the manager: appends across bucket boundaries, shallow and deep rollbacks, finalize-makes-pruned-gaps-canonical, snapshot stability across reorgs, and concurrent reads during writes.
- SMT and scheduler end-to-end tests: orphaned-version skipping, rollback, and finalize-past-an-orphan pruning (a finalized orphan must not reclaim its still-live predecessor).
- L1 bridge integration tests covering reorgs and the reorg filter.
- Full `./cargo-check-all.sh` and the workspace test suite pass.


biryukovmaxim (approved):

core/smt/src/updater.rs
```diff
@@ -25,11 +27,11 @@ impl<'a, S: Tree, W: WriteBatch> Updater<'a, S, W> {
         version: u64,
         mut commitments: Vec<Commitment>,
     ) -> [u8; 32] {
-        // Initialize the update context. Version > 0 is guaranteed by `Tree::update`.
-        let mut ctx = Self { tree, wb, prev_version: version - 1, version };
+        // Capture one snapshot for the whole update. Version > 0 is guaranteed by `Tree::update`.
+        let snapshot = tree.snapshot();
+        let mut ctx = Self { tree, snapshot, wb, prev_version: version - 1, version };
```
biryukovmaxim: prev version is likely fine here because of canonical checks in anoother place. but it makes sense to rename prev version and add comment why -1 is fine in that context (its only used as bound not as a real correct prev version

storage/canonical-chain/src/chain.rs
```diff
@@ -0,0 +1,149 @@
+use std::{
```
biryukovmaxim, line 26: Module-level doc for the write side: states what each operation does to the hot-zone/body layout, so the file orients on its own.

```suggestion
//! The single-writer build side of the canonical chain.
//!
//! Each mutation builds a brand-new [`CanonicalChainSnapshot`] and atomically publishes it, so a
//! reader holding an older snapshot keeps a stable view. See [`snapshot`](crate::snapshot) for the
//! bit layout (hot zone, body, finalized) that these operations maintain. In those terms:
//!
//! * [`append`](CanonicalChain::append) sets the next bit in the `tail` bucket. When the new id
//!   crosses into a fresh bucket the hot zone slides up by one: the old `tail` becomes the new
//!   `last_sealed`, the old `last_sealed` is sealed into the body ring, and a fresh empty `tail` is
//!   allocated (the sealing is done by [`HotZone::roll_forward`]).
//! * [`rollback`](CanonicalChain::rollback) clears every bit above `new_tip`. A shallow rollback
//!   rewrites only the copy-on-write hot buckets; a deep one that reaches a sealed bucket forks the
//!   body ring first, so earlier snapshots are untouched.
//! * [`finalize`](CanonicalChain::finalize) prunes the body buckets below the finalized id; once
//!   pruned, those buckets read as canonical.
//! * [`restore`](CanonicalChain::restore) rebuilds the whole layout from persisted ids at startup.

use std::{
```

storage/canonical-chain/src/chain.rs
```diff
@@ -0,0 +1,149 @@
...
+
+        // Publish the restored snapshot.
+        self.current.store(Arc::new(CanonicalChainSnapshot {
+            tip,
+            hot_zone: HotZone { tail_bucket, tail, last_sealed },
+            body: Arc::new(body),
+        }));
+    }
+
+    /// Marks `id` canonical as the new tip. Debug-panics unless `id` > tip.
```
biryukovmaxim, line 108: Document why setting the tail bit in place is safe for a reader still holding the previous snapshot (review item C1): the bit is for an `id` above that snapshot's `tip`, which `is_canonical` rejects first.

```suggestion
    /// Marks `id` canonical as the new tip. Debug-panics unless `id` > tip.
    ///
    /// The tail bit is set in place on a bucket the previous snapshot may still share, so a reader
    /// holding that older snapshot can observe the bit early. That is benign: the bit belongs to an
    /// `id` above the old snapshot's `tip`, which
    /// [`is_canonical`](CanonicalChainSnapshot::is_canonical) rejects before reading any bit.
```

storage/canonical-chain/src/chain.rs
```diff
@@ -0,0 +1,149 @@
...
+    pub(crate) fn append(&self, id: u64) {
+        // Read the current snapshot; the new id must extend its tip.
+        let cur = self.current.load();
+        debug_assert!(id > cur.tip, "append id {id} must extend tip {}", cur.tip);
+
+        // Monotonic append - reuse the shared ring and set the tail bit in place, no copy.
```
biryukovmaxim: Name the roll-forward/seal step in the append fast-path comment instead of just "monotonic append".

```suggestion
        // Roll the hot zone up to the new id's bucket (sealing any bucket left behind into the
        // body), then set the tail bit in place; the shared ring needs no copy for an append.
```

storage/canonical-chain/src/chain.rs
```diff
@@ -0,0 +1,149 @@
...
+        let (bucket, bit) = Bucket::locate(id);
+        let body = Arc::clone(&cur.body);
+        let hot_zone = cur.hot_zone.roll_forward(bucket, &body);
+        hot_zone.tail.set(bit);
+
+        // Publish the extended snapshot.
+        self.current.store(Arc::new(CanonicalChainSnapshot { tip: id, hot_zone, body }));
+    }
+
+    /// Rolls the chain back to `new_tip`, flipping off every id above it.
```
biryukovmaxim, line 124: Document the shallow-vs-deep rollback behavior and the body fork (review item R1).

```suggestion
    /// Rolls the chain back to `new_tip`, flipping off every id above it.
    ///
    /// A shallow rollback (one that stays within the hot zone) copy-on-writes only the hot buckets
    /// and shares the body ring. A deep rollback that reaches a sealed body bucket forks the ring
    /// first, so readers still holding the pre-rollback snapshot keep seeing the old bits.
```

storage/canonical-chain/src/hot_zone.rs
```diff
@@ -0,0 +1,60 @@
+use std::sync::Arc;
+
+use vprogs_core_atomics::AtomicRing;
+
+use crate::bucket::Bucket;
+
+/// The tail and last-sealed buckets, kept hot so a shallow reorg avoids the body ring.
```
biryukovmaxim: Give HotZone a role-level doc and a pointer to the layout diagram; the per-field details stay on the fields.

```suggestion
/// The top two buckets of the canonical chain, kept hot so an append or a shallow reorg never has
/// to touch the body ring. See [`snapshot`](crate::snapshot) for how the hot zone sits above the
/// body.
```

biryukovmaxim (commented):
Documentation suggestions for the canonical-chain crate: a data-model diagram (hot zone / body / finalized) on the snapshot, write-side operation docs on the chain, and one explanatory test. Also fixes the dangling `spec.md` reference (L1) and adds the C1/R1 rationale inline. Each item is a one-click suggestion.

storage/canonical-chain/src/hot_zone.rs
```diff
@@ -0,0 +1,60 @@
...
+    pub(crate) last_sealed: Arc<Bucket>,
+}
+
+impl HotZone {
+    /// The empty hot zone at bucket `0`.
+    pub(crate) fn empty() -> Self {
+        Self { tail_bucket: 0, tail: Arc::new(Bucket::new()), last_sealed: Arc::new(Bucket::new()) }
+    }
+
+    /// Advances to `target`, sealing buckets left behind into `body`, and returns the new hot zone.
```
biryukovmaxim, line 25: Explain how roll_forward slides the window up and seals displaced buckets (this is the "advancing" mechanic).

```suggestion
    /// Advances the hot zone to bucket `target`, sealing every bucket it leaves behind into `body`,
    /// and returns the new hot zone.
    ///
    /// Advancing slides the window up: the current `tail` and `last_sealed` (and any empty buckets
    /// between the old and new positions) are pushed into `body` in order, `target - 1` becomes the
    /// new `last_sealed`, and `target` gets a fresh empty `tail`. `target == tail_bucket` is the
    /// common no-op append case and reuses the current hot buckets unchanged.
```

storage/canonical-chain/src/lib.rs
```diff
@@ -0,0 +1,15 @@
+//! Fork-aware canonical chain over monotonic batch ids: a `block_hash`-keyed log with a lock-free
+//! `is_canonical(id)` overlay. See `spec.md` at the repo root for the full design.
```
biryukovmaxim: The crate doc points at a `spec.md` that doesn't exist on this branch (review item L1). Point at the in-crate docs instead.

```suggestion
//! `is_canonical(id)` overlay. See [`CanonicalChainSnapshot`] for the bit layout (hot zone, body,
//! finalized regions) and [`CanonicalChain`] for how appends, rollbacks, and finalization move it.
```

storage/canonical-chain/src/snapshot.rs
```diff
@@ -0,0 +1,50 @@
+use std::sync::Arc;
```
biryukovmaxim, line 35: Module-level data-model doc: an ASCII diagram of the finalized / body / hot-zone regions, how each reads, and a worked example. (Outer fence is 4 backticks so the inner ```text diagram survives.)

````suggestion
//! The immutable read side of the canonical chain.
//!
//! # Bit layout
//!
//! Every L2 batch owns a dense, 1-based, never-reused id, and the chain stores one bit per id:
//! `1` means canonical (on the tip's ancestry) and `0` means orphaned. Id `0` is the "no parent"
//! sentinel and is never canonical.
//!
//! Bits are grouped into fixed-size [`Bucket`]s of [`BUCKET_CAPACITY`](crate::BUCKET_CAPACITY)
//! consecutive ids: bucket `b` covers ids `b * CAPACITY + 1 ..= (b + 1) * CAPACITY`. A snapshot
//! splits those buckets into three regions, hottest on the right (`T` is `tail_bucket`, the highest
//! allocated bucket):
//!
//! ```text
//!   bucket:    0 ........... k-1 | k ........... T-2 | T-1 ......... T
//!              \__ finalized __/   \____ body ____/    \_ hot zone _/
//!               (pruned away)       (sealed ring)       last_sealed, tail
//!   read:       always canonical    the stored bit      the stored bit, set
//!               (bucket absent)     (true if pruned)    in place by the writer
//! ```
//!
//! * **Hot zone**: the top two buckets, held directly as `Arc<Bucket>`. `tail` (bucket `T`) is the
//!   one currently filling; `last_sealed` (bucket `T - 1`) is kept hot so any reorg that stays
//!   within the last two buckets touches only these and never the body ring. See [`HotZone`].
//! * **Body**: buckets `0 ..= T - 2`, sealed into an [`AtomicRing`]. A reorg deep enough to rewrite
//!   a sealed bucket forks the ring copy-on-write, so snapshots published earlier keep their view.
//! * **Finalized**: buckets below the body's base, pruned by `finalize` once their ids can no
//!   longer reorg. An absent bucket reads canonical, because every finalized id is on the final
//!   chain.
//!
//! For example, with `CAPACITY == 128` and `tip == 300`, id 300 sits at bucket 2 bit 43: `tail` is
//! bucket 2 (ids 257..=384), `last_sealed` is bucket 1 (ids 129..=256), and the body holds bucket 0
//! (ids 1..=128).

use std::sync::Arc;
````

File: storage/canonical-chain/tests/chain.rs [hunk 10]
```diff
@@ -0,0 +1,192 @@
+//! Canonical-bit behavior (appends, rollback, finalize, snapshots, copy-on-write) exercised the
+//! only way a consumer can reach it: by driving the [`CanonicalChainManager`].
+
+use std::{sync::Arc, thread};
+
+use vprogs_storage_canonical_chain::CanonicalChainManager;
```
biryukovmaxim: Import `BUCKET_CAPACITY` for the explanatory test below.

```suggestion
use vprogs_storage_canonical_chain::{BUCKET_CAPACITY, CanonicalChainManager};
```

storage/canonical-chain/tests/chain.rs
```diff
@@ -0,0 +1,192 @@
...
+    for i in 1..=10_000u64 {
+        manager.append(i);
+    }
+
+    assert!(is_canon(&manager, 1));
+    assert!(is_canon(&manager, 4_096));
+    assert!(is_canon(&manager, 4_097));
+    assert!(is_canon(&manager, 10_000));
+    assert!(!is_canon(&manager, 10_001));
+}
```
biryukovmaxim, line 78: Add an explanatory test: drive ids past two bucket boundaries and confirm a bit sealed into the body still reads correctly, using `BUCKET_CAPACITY` for exact boundaries.

```suggestion
}

#[test]
fn appends_slide_the_hot_zone_and_seal_older_buckets() {
    // Each bucket holds `BUCKET_CAPACITY` consecutive ids. The hot zone is the top two buckets
    // (`last_sealed`, `tail`); anything below them is sealed into the body. Filling three buckets'
    // worth of ids leaves bucket 0 in the body, bucket 1 as `last_sealed`, and bucket 2 as `tail`.
    let cap = BUCKET_CAPACITY;
    let mut manager = CanonicalChainManager::default();
    for i in 1..=3 * cap {
        manager.append(i);
    }

    // A bit sealed into the body still reads correctly through the snapshot.
    assert!(is_canon(&manager, 1), "first id, now in the body");
    assert!(is_canon(&manager, cap), "last id of body bucket 0");
    // The hot pair: bucket 1 (last_sealed) and bucket 2 (tail).
    assert!(is_canon(&manager, cap + 1), "first id of last_sealed bucket 1");
    assert!(is_canon(&manager, 2 * cap + 1), "first id of tail bucket 2");
    assert!(is_canon(&manager, 3 * cap), "tip, the last filled bit");
    assert!(!is_canon(&manager, 3 * cap + 1), "nothing above the tip is canonical");
}
```

biryukovmaxim:
Design note on the single-writer guard (`CanonicalChain::writer: Arc<AtomicBool>`):

Since the sole owner of the chain is statically known, the single-writer rule could be enforced at **compile time** rather than via the runtime `AtomicBool` + `claim_writer` panic:

- **(const) generic on the type** — encode the owner in `CanonicalChain`'s type so only the designated owner can construct the mutating handle; the type system rules out a second writer.
- **Capability token** — have the mutators (`append`/`rollback`/`finalize`/`restore`) take a singular, non-`Clone` writer token by value (move), so possession of the unique token *is* the write permission.

Either approach removes both the atomic and the `claim_writer` panic, turning a runtime assert into a static guarantee. Not blocking — just noting it as a possible follow-up.

core/smt/src/updater.rs
hmoog: I removed the prev_version field entirly then passing version - 1 as max_version to the node getter is self-explanatory.

storage/canonical-chain/src/chain.rs
hmoog: As discussed, this is already expressed in the inline comments and is not really relevant for the caller.

Anybody who wants to understand how the mechanism works under the hood will likely examine the code and see the inline comments for how the methods actually works.

storage/canonical-chain/src/hot_zone.rs
hmoog: Similar to the other reply. This is already explained in the inline comments and its irrelevant for a caller.

Anybody interested in the actual logic that achieves the contract will see the seal part as part if the inline comments.

storage/canonical-chain/src/chain.rs
hmoog: I actually disagree - the caller should not explain how the called method works but this should be explained in the called method itself (ideally.

If we ever refactor logic, then this needs to be kept in sync. From the perspective of the caller it is only relevant that the hotzone is advanced, that we share the body and set the bit in place.

So I updated to 

// Append - advance hot zone, reuse shared ring and set tail bit in place, no copy.

storage/canonical-chain/src/chain.rs
hmoog: I think this is an important invariant that should be spelled out - adding it.

storage/canonical-chain/src/hot_zone.rs
hmoog: Added the comment.

PR #85, opened 2026-06-18
Settle covenants in dev mode and under competing provers

biryukovmaxim:
# Settle covenants in dev mode and under competing provers

Stacked on **`use-receipt-storage`**. This PR delivers end-to-end covenant settlement
that runs **without a GPU** (dev mode) and **with more than one prover** settling the
same covenant, and that **survives a flaky L1 node** for a sustained multi-node run.

A `tn10-flow` node now bootstraps, resumes, or catches up to a covenant; proves bundles
on CPU under `RISC0_DEV_MODE`; settles each one on chain against a dev redeem variant
that needs no `OpZkPrecompile`; and reconciles with a competitor that lands a settlement
first by awaiting a single shared source of truth — the L1 bridge's settlement `watch`.
Transient wRPC timeouts in the settler, prover, and bridge are now bounded
retries instead of fatals, so one flaky remote node no longer kills the run. A committed,
env-driven runbook reproduces the two-node demo.

The change is split into the themes below.

## Dev-mode settlement

Lets the settlement worker land covenant settlements under `RISC0_DEV_MODE`, where the
prover emits stub receipts the production `OpZkPrecompile` would reject. A new
`SettlementMode { Production, Dev }` selects the redeem variant; the operating contract
is dev under `RISC0_DEV_MODE`, production only with real (CUDA) proofs.

- `covenant/build.rs`: `bootstrap_dev_covenant`, `dev_bootstrap_redeem`,
  `build_dev_settlement`, and the fixed `DEV_COVENANT_BUDGET` (the dev redeem has no
  precompile to size a budget for; a small fixed compute budget covers its hash + concat
  + equality ops). The dev path feeds the journal's `new_seq_commit` as the
  `claimed_seq_commit` the on-chain dev script `OpEqualVerify`s against
  `OpChainblockSeqCommit`, so guest-committed and chain values must agree. The dev redeem
  pins the output count to 1, so a bundle carrying L2→L1 exits is rejected here rather
  than silently dropped.
- `covenant/build.rs::build_settlement_for_mode` picks `build_settlement` vs
  `build_dev_settlement` off the settlement mode; the production settler and the sim both
  build through it.
- `tn10-flow` selects dev vs production by `dev_mode_enabled()` (`TN10_SETTLE=1` enters
  settlement mode); the sim `L2Driver` and `sim/tests/l2_flow.rs`
  (`l2_flow_proof_settlement_chain`, renamed from `…_real_proof_…`) run the full
  proving-driven settlement chain on CPU in dev mode and exercise `OpZkPrecompile` end
  to end only on the GPU box.

## Competing provers settling one covenant

Two provers can settle the same covenant concurrently; each forward-only-adopts the
competitor's on-chain settlement instead of wedging.

- `settle/remote.rs::classify_rejection` (inside `RpcSink`) maps a submit rejection by
  **which input** the node names: a fee-UTXO double-spend is `FeeRejected` (refund from
  another UTXO); the covenant outpoint already spent in the mempool is `Superseded` (drop
  the bundle, hold the covenant); an orphan names no input and is disambiguated by
  re-polling covenant liveness.
- `confirm.rs` `CovenantLiveness { Unspent, Spent, Shutdown }` and `covenant_liveness` (a
  single poll) tell a transient fee orphan from a competitor-confirmed covenant spend.
- `SettleOutcome { Advanced, Superseded, Shutdown }` (in `settle/settler.rs`) replaces the
  old `Option<CovenantState>`.
- `submit_jitter` (a `Range<u64>` ms window carried by `RpcSink`) breaks the deterministic
  spend-race so the winner alternates rather than one prover sweeping every range; `None`
  in production.

## Proving / settlement decoupling (livelock fix)

Previously the aggregate prover folded external settlements into a settled lower bound
and dropped/superseded bundles, which could livelock two provers. Now proving and
settlement ordering are independent.

- `aggregate-prover/src/worker.rs`: the prover proves **and forwards every contiguous
  bundle**. Removed `absorb_external_settlements`, `drop_settled_prefix`, the `from_block`
  / `from_state` bookkeeping, and the `prev_state == from_state` debug assert. The
  `covenant_id` and `new_seq_commit` asserts (the real correctness guards) stay.
- `scheduled_bundle.rs`: dropped `latest_settlement` from `ScheduledBundle` and its
  constructors — bundles no longer carry settlement state.
- `l1/bridge/src/worker.rs`: the bridge publishes the tip's canonical `last_settlement`
  into a `tokio::sync::watch` channel (`publish_settlement` `send_replace`s after seed and
  after every batch). The bridge is the single writer (the `Sender`).
- `settler/src/worker.rs`: the settler **reads that watch** (a `Receiver`) to decide
  settle / skip / adopt. Adoption is gated forward-only: the value must be *ahead*
  (`s.new_state != cov.state`) and not behind (`s.daa_score >= cov.daa_score`); a value
  behind `cov` (a competitor already passed) is ignored. Re-enables the two contention
  tests.
- `bundle_size: RangeInclusive<usize>` knob on `AggregateProverConfig` (min ready before
  a bundle forms, max per bundle); `1..=usize::MAX` is the greedy default, asserted
  non-empty at startup.
- `SettlementInfo` gains `daa_score` (unaligned little-endian `U64` so the struct stays
  `Unaligned`); Borsh is hand-rolled with a round-trip guard test. `settlement_info`
  takes the containing block's DAA score so the published settlement carries it.

### Known follow-up — re-form superseded ranges ([#88](https://github.com/kaspanet/vprogs/issues/88))

This decoupling removes the *absorb-external-settlements* livelock, but a **distinct,
pre-existing** one remains and is **intentionally not addressed here**: when a competitor
settles a **shorter** range than this prover proved, the orphaned tail is dropped, not
re-formed against the adopted tip, so two provers with divergent bundle boundaries can
wedge once fresh activity stops. The in-tree `TODO(livelock)` at
`examples/tn10-flow/tests/two_provers_contend.rs` documents it, and the contention tests
sidestep it by injecting fresh ranges during drain rather than asserting an acceptance-only
drain.

This PR is already large, and the fix (retain consumed batches and re-aggregate the
suffix from the adopted state — feasible without re-proving, since each batch journal
already binds its own `prev_state`) is **orthogonal** to the dev-mode / contention /
resume work landed here. It is therefore tracked separately in **#88** and is best
resolved in a follow-up **stacked on top of this PR**, which also restores the
acceptance-only drain assertion.

## Settler tip from the live handle (RPC-flood fix)

The settler's old startup tip resolution scanned the **whole virtual chain**
(`get_virtual_chain_from_block_v2`) on every re-resolve, flooding the L1 node and
panicking on a transient timeout.

- `settler/src/worker.rs::run` now establishes the starting covenant tip **solely from
  the settlement `watch`** — the same `last_settlement` a chain scan would derive, with
  zero scan RPC. When the watch is empty and there is no `start_from` (a fresh deploy with
  an unspent bootstrap), it confirms the bootstrap UTXO directly; otherwise it leaves `cov`
  at the bootstrap and lets mid-stream adoption advance it.
- The whole-chain tip-scan path and the intermediate `worker/resolve.rs` module were
  added and then **removed entirely**.

## Network selection

- `TN10_NETWORK` selects the Kaspa network: `devnet` / `simnet` / `mainnet` /
  `testnet-N` (any suffix) / `testnet-10`|`testnet10`|`tn10`, defaulting to testnet-10.
  Parsed in `config.rs` (panics on an unrecognized value). `main.rs` uses
  `cfg.network_id` instead of a hardcoded testnet-10; `L1Node::new` takes a `NetworkId`.

## Resume and catch-up

A settlement node can resume a persisted covenant or catch up to an existing one from an
empty db, replaying L1 forward from the covenant's deploy block.

- `l1/bridge`: `start_from: Option<Hash>` and `with_start_from` / `seed_from_block` seed
  the bridge's fresh-chain root at an explicit deploy block. Fresh-chain precedence:
  `start_from` > `seed_depth` > pruning point.
- `tn10-flow/src/persistence.rs`: persists `bootstrap_block_hash` (the deploy block) and
  exposes `bootstrap_block()` / `bootstrap_txid()`. A missing field reads as `None`
  (back-compat).
- `tn10-flow/src/main.rs::start_settlement` splits the cases:
  - **Resume** — a populated data dir reconstructs the same covenant from persisted
    identity (`covenant_hash().or(env)`) instead of re-bootstrapping and clobbering the
    state file.
  - **Catch-up** — an env-supplied `TN10_COVENANT_ID` joins an existing covenant,
    reconstructing the never-advanced bootstrap state; the bridge replays from
    `TN10_START_FROM` and the settler self-heals from the on-chain `last_settlement`.
  - **Bootstrap** — a fresh deploy at the empty SMT; the captured sink is persisted as
    the deploy block.
  - **Guard:** a fresh catch-up with no `TN10_START_FROM` **fails fast** — seeding only
    `seed_depth` below the sink loses pre-seed lane history and corrupts the
    reconstructed `seq_commit` (the aggregate-prover's `new_seq_commit` assertion stays a
    correctness guard; the bad input is prevented, not the guard weakened).
  - `TN10_BOOTSTRAP_TXID` is optional: a never-settled covenant confirms its bootstrap
    UTXO; one already settled past its bootstrap is adopted from `last_settlement` (the
    spent bootstrap UTXO is useless), so no txid is needed there.

## RPC resilience

Transient wRPC timeouts against a live testnet node are now bounded retries instead of
panic/fatal; terminal errors stay fatal.

- `tn10-flow/src/daemon.rs` `RemoteLaneSource::fetch_lane_proof`: the prover's lane-proof
  RPC retries (`LANE_PROOF_MAX_ATTEMPTS = 10`, `500ms`) instead of `.expect`-panicking the
  prover worker.
- `l1/bridge/src/worker.rs` `get_vcc_with_retry`: the chain-init backfill and steady-state
  follow retry transient `Error::Rpc` (`RPC_RETRY_MAX_ATTEMPTS = 10`, `500ms`,
  shutdown-interruptible); terminal errors (`CheckpointLost`, `BackfillTargetNotFound`)
  return immediately so a real reorg/prune past the root still fatals.

## Runbook

- `examples/tn10-flow/scripts/` (`run-demo.sh`, `monitor.sh`, `README.md`): a committed,
  env-driven two-node settlement demo. Node A bootstraps a covenant; node B catches up to
  the same id; both settle in lock-step; the monitor reports per-node health and a
  cross-node consistency check. **No node address or private key is embedded** — the
  scripts read `TN10_WRPC_URL`, `TN10_KEY1`, `TN10_KEY2` from the environment and error
  if unset. Documents the three start modes (bootstrap / resume / catch-up).

## Generic trait-based settler + notification-based confirmation

One settlement component drives both the production daemon and the L2 sim through the
same build → fund → submit → await-confirm → advance flow, injecting per-environment
behavior through two traits. Confirmation **awaits a notification** instead of polling the
node's utxoindex. This generalizes the `build_settlement_for_mode` sharing into a full
loop and supersedes the earlier handle/polling mechanism.

- **Notification handle.** The settlement publish target is a `tokio::sync::watch` channel
  (replacing `Arc<ArcSwapOption<SettlementInfo>>`): the bridge holds the `Sender`; each
  settler holds a `Receiver` (one per settler — `two_provers_contend` has two off one
  sender), `borrow()`ing it for resume/adoption and **awaiting** it (`wait_for`) for
  confirmation. `SettlementWorkerConfig.settlement` is a required `watch::Receiver`.
- **Two traits + generic `Settler`** (`settler/src/settle/{effects,settler,remote}.rs`):
  `FeeSource::fund` and `SettlementSink::submit -> SubmitOutcome { Accepted, FeeRejected,
  Superseded, Fatal, Shutdown }` (RPITIT, no `Box`/`async-trait`, matching
  `LaneProofSource`). `Settler<F, K>::settle_one` builds, funds/submits with the fee-retry
  loop, then confirms by awaiting the watch, folding confirmation and competitor-supersede
  into one await — nothing submits-then-polls. The production daemon drives it with
  `WalletFeeSource`/`RpcSink`.
- **Residual RPC confirm.** `confirm.rs` (a top-level `pub(crate)` module) keeps
  `confirm_outpoint`/`covenant_liveness` only for the two cases with no watch event: the
  one-time fresh-deploy bootstrap confirm and the orphan-liveness poll. Steady-state
  confirmation is fully notification-based.
- **Sim.** `L2Driver` funds settlements through a `SimFeeSource: FeeSource` (the in-memory
  counterpart to `WalletFeeSource`) on top of the shared `build_settlement_for_mode`,
  while keeping its deterministic `observe_covenant` confirmation and remaining its own
  miner (no detached settler task) — the choice that preserves `l2_flow_is_deterministic`.
- `worker.rs` is the production driver: it builds a `Settler` from the real impls and runs
  the queue loop (pop, await artifact, reconcile/adopt, settle). `worker/config.rs`
  retains `SettlementWorkerConfig`/`SettlementMode`/`AlternationPacer`.

## Test-only pacer, plumbing

- `AlternationPacer` (forces two competing settlers to alternate so neither sweeps every
  range) is **gated behind a `test-utils` feature** and compiled out of production;
  `SettlementWorkerConfig` carries no `alternation` field in a normal build.
  `cargo-check-all.sh` adds a `--features test-utils` pass so the gated code can't rot.
- New supporting types: `OutpointAt` (`pub`), `Elfs`, `BridgeObservers`.
- Wallet `pay_to_address` / `pay_to_address_transaction` and `L1Node::fund_address` seed
  a distinct prover's fee key from a coinbase wallet (used by the contention tests).
- Adapts to the upstream `fix-covenant-binding` rusty-kaspa branch
  (`set_storage_mass`, `compute_commit`, `Lane`/`proof.lane`, `with_covenant` change
  output); excluded-crate lockfiles resynced; debugging cleanup.

## Tests

- `examples/tn10-flow/tests/two_provers_contend.rs` (builds only with
  `--features test-utils`): `two_provers_contend`, `prover_catches_up_to_existing_covenant`,
  `prover_catches_up_to_already_settled_covenant`, `prover_resumes_after_settlement`,
  `prover_resumes_after_settlement_contended`.
  All five pass under `RISC0_DEV_MODE=1`, exercising the watch-based confirm and the
  supersede / adopt / resume paths through the generic `Settler`.
- `sim/tests/l2_flow.rs::l2_flow_proof_settlement_chain` runs the proving + settlement
  chain (through `SimFeeSource` + `build_settlement_for_mode`) off the same operating
  contract; the full `l2_flow` suite incl. `l2_flow_is_deterministic` passes in dev mode.
- `SettlementInfo` Borsh round-trip guard; `settlement_info` DAA-score assertions.


l1/bridge/src/config.rs
```diff
@@ -30,9 +31,19 @@ pub struct L1BridgeConfig {
...
+    /// pruning point. `None` defers to the lower-precedence options.
+    pub start_from: Option<Hash>,
     /// Optional observer the bridge publishes its latest chain-block DAA score into, for an
     /// external progress reporter; `None` disables publishing.
     pub tip_daa: Option<Arc<AtomicU64>>,
+    /// Optional `watch` sender the bridge publishes the tip's last covenant settlement into. The
+    /// bridge is the single writer; each settler holds a [`watch::Receiver`] it borrows (and, once
+    /// confirmation is notification-based, awaits) to reconcile against the canonical settlement
+    /// without a confirm RTT. `None` disables publishing.
+    pub settlement: Option<watch::Sender<Option<SettlementInfo>>>,
```
hmoog: Maybe we should rename to settlement_observer to make it more self-explanatory since "settlement" alone is pretty ambiguous.

zk/backend/risc0/settler/src/confirm.rs
```diff
@@ -0,0 +1,126 @@
...
+    max_polls: u32,
+) -> ConfirmOutcome {
+    let prefix = Prefix::from(params.net.network_type());
+    let address =
+        extract_script_pub_key_address(target.spk, prefix).expect("covenant P2SH address");
+    for _ in 0..max_polls {
+        if shutdown.is_open() {
+            return ConfirmOutcome::Shutdown;
+        }
+        let utxos = client
```
hmoog, line 72: In the bridge, the RPC connections use retry semantics but here we don't retry, so a single e.g. timeout crashes the settler.

Should we maybe generalize the retry semantics into a generic method so we can reuse it across the code-base or is this meant to be a quick and dirty implementation for the MVP that gets hardened later?

zk/backend/risc0/settler/src/settle/settler.rs
```diff
@@ -0,0 +1,160 @@
...
+        let mut excluded = HashSet::new();
+        let txid = loop {
+            // Re-funding from another UTXO each rejection is the one unbounded wait here; bail on
+            // shutdown so teardown is not held up retrying a doomed submission to exhaustion.
+            if shutdown.is_open() {
+                return SettleOutcome::Shutdown;
+            }
+            let Some(funded) = self.funder.fund(&built, covenant_entry.clone(), &excluded).await
+            else {
+                panic!("settlement submit failed: every spendable fee UTXO was rejected");
```
hmoog: Should this be recoverable?

zk/aggregate-prover/src/worker.rs
```diff
@@ -136,34 +142,37 @@ where
...
             take += 1;
         }
+        // Park until at least the configured minimum are consecutively ready. Returning false
+        // leaves the batches queued; the run loop re-tries when more arrive. The extend
+        // loop above caps `take` at the maximum, so a `take` short of the minimum is
+        // genuinely short of ready batches, not the cap. With the default `1..`, `take >= 1
+        // == *start()` always, so this never fires and behavior is identical to
+        // greedy-from-1.
+        if take < *self.bundle_size.start() {
+            return false;
```
hmoog, line 167: The return false here parks the run loop, but nothing re-wakes it on a receipt publication — run()'s select! only listens on the inbox (batch/rollback) and shutdown. With start() > 1, a tail shorter than the minimum  (or one whose later receipts publish after the last command) strands permanently at an idle chain tip. Invisible with the default 1.. but reachable via the 2..=4 config in two_provers_contend.rs. 

Suggest also awaiting the next unpublished batch's latch in the park select! so the park self-heals.

sim/src/driver/l2_driver.rs
```diff
@@ -203,17 +216,23 @@ fn build_exec(
...
             backend.clone(),
             store.clone(),
+            state.receipt_store(),
             BatchProverConfig { lane_key: lane, covenant_id },
             AggregateProverConfig {
                 lane_key: lane,
                 covenant_id,
                 lane_source: ConsensusLaneSource::from_weak(consensus),
                 settlement_queue: Some(queue.clone()),
+                bundle_size: 1..=usize::MAX,
```
hmoog: bundle_size is hardcoded here, so the driver's configured bundle_size (set at :284, l2_flow.rs passes 10) never reaches the prover - it bundles greedily size-1, not in tens. The config value only gates settlement      back-pressure at :685/:708.

The l2_flow comment "one proof per bundle_size blocks" is therefore false; that path is untested. Thread the field into build_exec, and consider splitting the two meanings (proof granularity vs. drain threshold) since one name covering both is what hid this.

hmoog (approved):
Okay here are a few remarks but as we discussed in telegram maybe some can be deferred for later.

Generally speaking I think we can harden things later and maybe make progress with just merging things as most of the things that I found are outside of the core framework.

You decide what is worth addressing before a merge.

zk/backend/risc0/settler/src/confirm.rs
biryukovmaxim: Id rather unify them behind one retry abstraction in dedicated followup. let keep the as is currently

l1/bridge/src/config.rs
biryukovmaxim: renamed

zk/backend/risc0/settler/src/settle/settler.rs
biryukovmaxim: https://github.com/kaspanet/vprogs/pull/85/changes/02a6b765c14ce3b0e4e3df73e5e6c624b643d3c1
fixed

zk/aggregate-prover/src/worker.rs
biryukovmaxim: https://github.com/kaspanet/vprogs/pull/85/changes/e816d352d0214beb60946233ba4824942dae975b

sim/src/driver/l2_driver.rs
biryukovmaxim: https://github.com/kaspanet/vprogs/pull/85/changes/4e775bccbbcbf391e4bcc72f5c8bfd056cb4cfca

zk/aggregate-prover/src/worker.rs
hmoog: Looks good

PR #86, opened 2026-06-18
Feat: SMT aware fork (backup)

hmoog:
This design is really ugly - planning to move the CanonicalChain as a versioning concept directly into the Store but this will be a larger refactor.

Just storing this here so the changes dont get lost.


hmoog:
superseded by the fork aware Canonical Chain tracking.

PR #90, opened 2026-06-24
Re-aggregate a superseded suffix in the aggregate prover

### LINKED ISSUE #88: Aggregate prover does not re-form a superseded range after a shorter competitor settlement
## Summary

When a competitor settles a **shorter** range than this prover has already proved, the proved-but-unsettled tail is **dropped, never re-formed**. The aggregate prover consumes its batches into one coarse bundle (e.g. `[1..10]`, `prev_state = S0`); if a competitor lands a shorter settlement first (e.g. `[1..5]`, advancing the covenant to `S5`), this prover's `[1..10]` bundle no longer chains (`prev_state S0 != S5`) and is skipped. Nothing re-aggregates the residual `[6..10]` against the adopted tip `S5`, so blocks 6–10 stay proved-but-unsettleable. With two contending provers whose bundle boundaries diverge, the continuation chain can **wedge once fresh activity stops**.

This is already known internally — see the `TODO(livelock)` at `examples/tn10-flow/tests/two_provers_contend.rs:520-527`:

> *the aggregate prover currently has one [livelock]: a proved-but-unsettled bundle that a competitor supersedes is dropped, not re-formed against the adopted tip, so two contending provers can wedge once fresh activity stops. ... Restore the acceptance-only assertion once the aggregate prover re-forms superseded ranges.*

## Root cause

1. **Batches are consumed and not retained.** `zk/aggregate-prover/src/worker.rs:164-165` pops the bundle's batches off the queue:
   ```rust
   let bundle: Vec<ScheduledBatch<S, P>> =
       (0..take).map(|_| self.queued.pop_front().unwrap()).collect();
   ```
   Once the bundle is published, the batches (and their per-batch receipts) are not held anywhere.

2. **No feedback path from settlement back to the prover.** The aggregate prover only handles `Command::Batch` / `Command::Rollback` (`zk/aggregate-prover/src/worker.rs:100-103`). The settler, on supersession, simply skips and continues (`zk/backend/risc0/settler/src/worker.rs:166-171`); it never tells the prover "the on-chain tip is now `S5`, re-prove from there." Bundle formation is forward-only, driven solely by newly-arriving batches.

3. **The settler can adopt the tip but the prover can't act on it.** The settler adopts the competitor's `SettlementInfo` (`worker.rs:143-165`) and skips the superseded bundle, then waits on `queue.wait_and_pop()` for the *next* bundle — which, after the prover has consumed those batches, never covers the orphaned `[6..10]` range.

## Why the fix is feasible (re-aggregation, not re-proving)

Re-forming the suffix does **not** require re-running the expensive per-batch proofs. The aggregator anchors a bundle's `prev_state` to its **first batch** and chains each subsequent batch by equality:

- `zk/abi/src/batch_aggregator/verifier.rs:98` — `prev_state: &self.first_batch.prev_state`
- `zk/abi/src/batch_aggregator/verifier.rs:81` — `assert_eq!(this.prev_state, prev.new_state, "prev_state")`
- output `new_state` = last batch's `new_state` (`verifier.rs:100`)

Each per-batch journal already carries its own `prev_state`/`new_state` (`zk/abi/src/batch_processor/journal/batch_transition.rs`). The batch for block 6 intrinsically binds `prev_state = S5`. So re-aggregating the suffix `[6..10]` (first batch = block 6) yields a valid `S5 -> S10` settlement, reusing the existing per-batch receipts as composition assumptions — **only the cheap aggregator proof reruns** (and the final-block lane proof is unchanged, since `block_prove_to` is the same).

## Proposed fix

- **Retain** consumed batches (or their per-batch receipts + journals) in the aggregate prover until a settlement covering their range confirms. The settlement watch (`SettlementInfo { new_state, block_prove_to, .. }`) is the signal that a range is durably settled.
- **Add a settlement-aware feedback path** to the prover (a new `Command`, or have the prover observe the same settlement watch the bridge writes). On a confirmed tip `ST` that does not match the prover's last-formed bundle, drop retained batches at/below `ST` and **re-aggregate the suffix** beyond `ST` into a fresh bundle whose first batch's `prev_state == ST`.

This dovetails with the existing TODOs at `zk/backend/risc0/settler/src/worker.rs:104-111` ("track which settlements are done vs pending and persist that").

## Test design (acceptance-only drain)

The current contention tests mask the gap by **injecting fresh ranges** during the drain phase, which always hands the trailing prover new batches that eventually realign. The regression this needs is the assertion the `TODO(livelock)` says to restore:

1. Reuse the `two_provers_contend` harness, but give the two provers **divergent bundle sizes** (e.g. A `5..=5`, B `3..=3`) so their range boundaries deliberately misalign.
2. Drive a handful of contended ranges normally (fresh carriers + acceptance), so both provers prove overlapping ranges and at least one coarse bundle gets superseded and dropped.
3. **Stop injecting fresh ranges.** Run an **acceptance-only drain**: mine only acceptance blocks for a *bounded* number of rounds (so the test fails by assertion rather than hangs).
4. Assert the covenant continuation chain advances to settle **every block both provers have proved through** (chain tip reaches the highest proved block / chain length reaches the expected total).

Today step 4 fails: the chain stalls below the highest proved block because the orphaned suffix is never re-formed. Once the prover re-forms superseded ranges, the acceptance-only drain advances to full coverage and the assertion at `two_provers_contend.rs:520-527` can be restored.

## Impact

- Proved work for the orphaned tail is wasted and, in the worst case, the continuation chain wedges with no fresh activity to bail it out — defeating the multi-prover redundancy the contention path is meant to provide (if the prover holding a usable continuation dies, the survivor cannot take over its already-proved range).
- Single-operator / steady-state-then-idle deployments are exposed: once L2 activity quiesces, a partially-superseded range never settles.

biryukovmaxim:
Stacked on #85 (`devmode-settlement`). This PR adds a single commit on top of that branch; review the delta with `git diff devmode-settlement..HEAD`. The description below covers only that delta.

## Problem

When a competitor settles a **shorter** range than this prover has already proved, the proved-but-unsettled tail was dropped and never re-formed. The aggregate prover consumed its batches into one coarse bundle (e.g. `[1..10]` from `S0`); a shorter competitor settlement (e.g. `[1..5]`, advancing the covenant to `S5`) left that bundle no longer chaining (`prev_state S0 != S5`), so the settler skipped it. Nothing re-aggregated the residual suffix `[6..10]` against the adopted tip `S5`, so those blocks stayed proved-but-unsettleable, and two contending provers whose bundle boundaries diverge could wedge once fresh activity stopped.

## Change

The aggregate prover now **retains** the batches it consumes and **re-aggregates the surviving suffix** against the adopted tip, so two contending provers converge on one continuation chain regardless of who settles first or how batches group into bundles.

- The prover observes the bridge's `last_settlement` watch. On a settlement landing on a retained block boundary it drains that prefix and re-forms the remainder into a fresh bundle whose first batch already chains from the adopted tip. The cached per-batch receipts are reused as composition assumptions, so only the cheap aggregator proof re-runs (no per-batch re-proving).
- An **unmatched** settlement boundary drops nothing (never clears the retained window), so an unsettled suffix is never lost.
- Rollback truncates the retained buffer alongside the queued one.
- The watch receiver is plumbed through `AggregateProverConfig` / `ProvingParams`; the sim opts out (`None`).

## Tests

- `settled_prefix` unit tests in `vprogs-zk-aggregate-prover` deterministically cover the drain decision, including the must-not-clear-on-unmatched-boundary case.
- `two_provers_reform_superseded_suffix`: two provers with divergent bundle sizes (`5..=5` / `3..=3`) contend over one covenant, then an acceptance-only drain confirms they converge on a single non-forked continuation chain that keeps advancing, with neither settler panicking.

Closes #88


hmoog (approved):
LGTM

PR #95, opened 2026-06-30
Deposit/transfer/withdraw runtime flow test (+ self-authorizing config Init)

biryukovmaxim:
> Stacks on `runtime-poc-continue` (not yet in `master`), so most of the diff against `master` is that branch. Described below are **only the changes on top of `runtime-poc-continue`** (2 commits).

## Deposit/transfer/withdraw runtime flow test

A deterministic harness verifying the real `runtime-processor` guest end-to-end — from carrier-tx construction through execution, two-node consensus, and settlement.

- **Reusable builders** (`zk/backend/risc0/test-suite/src/runtime_flow.rs`): deterministic `RuntimeSigner` keys and a `RuntimeCarrier` that emits signed deposit / transfer / transfer-create / withdraw / init / update / lock-rotation carriers, driving both unfunded scheduler txs and mempool-funded carriers.
- **Test binaries** (run under `RISC0_DEV_MODE=1`):
  - `runtime_flow_e2e` — every action through `Scheduler` + `Vm` + the runtime ELF; transfer-before-deposit rejection; deposit→transfer within one block.
  - `runtime_flow_l1_e2e` — **two** framework nodes following one real kaspad simnet via their bridges, agreeing on the full flow including out-of-order rejection.
  - `runtime_flow_settlement_e2e` — deposit + withdraw through batch proving + aggregation + settlement; asserts the deposit-address commitment, a non-zero permission-tree exit commitment, and the count==2 settlement structure.
  - `runtime_flow_scenario` — an expected-state model cross-checked against the runtime, with a seeded random generator (validated to 300 mixed actions: 279 accepted / 21 rejected, model matched the runtime on every one). Scale via `FLOW_SCENARIO_ACTIONS` / `FLOW_SCENARIO_USERS` / `FLOW_SCENARIO_SEED`.
- **Supporting changes**: `runtime_processor_elf()` loader in the test-suite; a generic closure-based `Wallet::build_signed_carrier` in `l1/wallet` for funded carriers; public `L1Node::wallet()`.

The on-chain settlement landing (`OpZkPrecompile` script check) needs a real CUDA proof and stays gated behind `!dev_mode_enabled()`, matching the repo's real-proving policy; everything else (execution + settlement construction) is verified in dev mode.

```
RISC0_DEV_MODE=1 cargo test -p vprogs-zk-backend-risc0-test-suite \
  --test runtime_flow_e2e --test runtime_flow_l1_e2e \
  --test runtime_flow_settlement_e2e --test runtime_flow_scenario -- --test-threads=1
```

## Fix: config `Init` is self-authorizing

The genesis-gated `Init` action builds an ad-hoc genesis lock and checks it against the auth context, but the only signers that could populate that entry (`SchnorrSigPtr`) derive the pubkey from the **target resource's lock** — and the config resource is empty at `Init`, so resolution failed with `"unknown or absent resource kind"`. `Init` could previously only be authorized by spending a genesis-keyed P2PK output via `PrevTxV1Witness`.

Added `GenesisSchnorrSigPtrSigner` (kind `0x05`): same `u32 sig_offset` body as the regular Schnorr signer, but it verifies the signature against the baked-in `GENESIS_SCHNORR_BYTES` instead of reading the resource lock, so `Init` is a self-contained genesis-signed action (the witness/funded-init workaround in the flow harness was removed).

Guest ELF rebuilt via `./zk/backend/risc0/build-guests.sh runtime-processor`.


l1/wallet/src/build.rs
```diff
@@ -220,6 +220,104 @@ pub fn covenant_bootstrap_transaction(
...
+            payload,
+        );
+        let signed = sign(MutableTransaction::with_entries(tx, entries.clone()), args.keypair).tx;
+        commit_storage_mass(args.params, &signed, &entries);
+        signed
+    };
+
+    // The payload length is independent of the rest_preimage, so an empty-rest probe sizes the tx.
+    let probe_payload = (args.finalize_payload)(&[]);
+    let probe = build(0, probe_payload.clone());
```
hmoog, line 292: **The zero-fee probe runs before the funding-sufficiency assert, so an oversized `extra_outputs` underflows instead of hitting the intended error.**

`build(0, probe_payload)` evaluates `args.entry.amount - extra_value - fee` (line 267). When `extra_value > args.entry.amount`, that subtraction underflows: a panic in debug, or in release a wrap to a ~1.8e19-sompi change output that then gets built and signed into a nonsense transaction. The assert on line 288 that would have reported this clearly never gets the chance.

This is reachable: `Wallet::build_signed_carrier` picks `utxos.into_iter().next()`  (an arbitrary UTXO, not one verified large enough for the deposit output it's asked to fund).

The sibling `pay_to_address_transaction` gets this right: it asserts `args.entry.amount > payout` on line 142, *before* its first `build`, so the guard was dropped here rather than never existing. Please hoist an `extra_value`-vs-`amount` check above the probe.

zk/backend/risc0/test-suite/src/runtime_flow.rs
```diff
@@ -0,0 +1,552 @@
...
+
+    /// The kaspa P2PK address for this key under `prefix`.
+    pub fn p2pk_address(&self, prefix: Prefix) -> Address {
+        Address::new(prefix, Version::PubKey, &self.pubkey)
+    }
+
+    /// The genesis authority key (secp256k1 scalar `3`), the only key allowed to `Init` the config.
+    pub fn genesis() -> Self {
+        let s = Self::from_scalar(3);
+        debug_assert_eq!(
```
hmoog: **`debug_assert_eq!` is compiled out under `cargo test --release`, so the scalar-3/genesis-pubkey correspondence is unchecked exactly when it breaks.**

`genesis.rs` carries a live TODO to replace the baked-in pubkey with a real ceremony output before mainnet. The moment that lands, scalar 3 no longer matches `GENESIS_SCHNORR_BYTES`. In a release test run this assert vanishes, so the harness builds `Init` carriers signed by the wrong key and every flow test fails deep inside the guest with `init: not authorized by genesis pubkey` (instead of the one clear message this assert was written to produce).

This is the same defect class as in the previous review round (`debug_assert` standing in for a check that was meant to be real). It runs once per test, so a hard `assert_eq!` costs nothing.

zk/backend/risc0/runtime-processor/src/signer_variants.rs
```diff
@@ -107,6 +108,41 @@ impl<'a> Signer<'a> for PrevTxV1WitnessSigner {
...
+impl<'a> Signer<'a> for GenesisSchnorrSigPtrSigner {
+    const TAG: u8 = 0x05;
+    type Unlocker = SchnorrUnlocker;
+
+    fn decode(buf: &mut &'a [u8]) -> CodecResult<Self> {
+        let sig_offset = buf.le_u32("signer.genesis_schnorr.sig_offset")?;
+        Ok(Self { sig_offset })
+    }
+
+    fn resolve(
```
hmoog, line 133: **`resolve` ignores `resource_idx` entirely: it's the first signature signer that doesn't bounds-check its target against the resource list.**

Every other signature-flavoured signer reaches `ctx.resources.get(resource_idx)` via `read_schnorr_lock_pubkey` / `read_multisig_lock_pubkey_at` and errors on an out-of-range index. This one never touches `ctx.resources`, so a signer entry naming `resource_idx = 200` in a 2-resource tx resolves cleanly and pushes `(200, SchnorrUnlocker{GENESIS})` into `auth.schnorr`.

Not exploitable today: `decode_ix` order-checks signer indices, an entry at a bogus idx can never match a real action's `updater_idx`, and forging one needs the genesis key anyway. But the previous review's finding explicitly rested its fail-closed argument on *"`resources.get(idx)` errors for signature signers; witness signers ignore the idx"*. This signer quietly moves a signature signer into the "ignores the idx" camp and weakens that argument. A `ctx.resources.get(resource_idx as usize).ok_or(...)?` restores the invariant for the cost of one line.

Separately, on the doc above (lines 111-119): the framing *"for the bootstrap `Init` action"* is aspirational, since nothing in `resolve` ties this signer to `Init` or to the config resource. The restriction is emergent: `apply_init` is simply the only action that authorizes against a resource with no readable lock. Worth saying that explicitly, since it's load-bearing and non-obvious.

zk/backend/risc0/test-suite/tests/runtime_flow_scenario.rs
```diff
@@ -0,0 +1,378 @@
...
+        ExecutionConfig::default().with_processor(vm),
+        StorageConfig::default().with_store(storage.clone()),
+    );
+
+    let mut model = Model::new();
+    let mut accepted_count = 0usize;
+    let verbose = std::env::var("FLOW_SCENARIO_VERBOSE").is_ok();
+    let started = Instant::now();
+    for (i, action) in actions.iter().enumerate() {
+        let accepted = model.predict(action);
```
hmoog, line 224: **The model's two most interesting predictions (the per-action accept/reject verdict and the exit count) are computed and then only printed, never asserted.**

`model.predict(action)` returns a verdict that feeds only `accepted_count` (an `eprintln` on line 248) and an optional verbose log (line 232). `model.exits` is incremented on line 148 and likewise only printed (line 259). The sole assertion in the whole engine is the end-state balance cross-check on line 244. Consequences:

- **Accept/reject is unverified per-action.** An action the runtime accepts but the model rejects (or vice versa) passes undetected whenever the net balance effect coincides, which is precisely the premature/out-of-order cases this engine exists to test.
- **Exits are invisible.** A `Withdraw` emitting zero exits, two exits, or one with the wrong SPK or amount all pass; only the balance debit is checked. `model.exits` is right there and asserts nothing.
- **Config state is never read back.** `config_min_withdrawal` is tracked in the model, and `config_min_withdrawal()` is exported from the harness for exactly this, but the config resource is never read. `InitConfig`/`UpdateConfig` are only verified indirectly, via later withdraw behaviour.

This is the bundle-level analogue of a finding from an earlier round. Asserting `accepted` against the runtime per action, plus the exit count and the config min at the end, would make the engine test what its doc comment claims it tests.

zk/backend/risc0/test-suite/tests/runtime_flow_e2e.rs
```diff
@@ -0,0 +1,212 @@
...
+    let alice = RuntimeSigner::user(30);
+    let alice_new = RuntimeSigner::user(31);
+
+    let mut h = Harness::new();
+    h.commit(init_config_tx(100, cov, &owner));
+    h.commit(deposit_tx(cov, &alice, 5_000));
+    assert_eq!(h.balance(&alice), Some(5_000));
+
+    // Rotate alice's controlling key to alice_new. The resource stays at alice.user_id().
+    h.commit(rotate_user_lock_tx(&alice, &alice_new));
```
hmoog, line 168: **`rotate_user_lock_preserves_balance` would pass against a completely no-op `UpdateUserLock`.**

The test rotates alice's lock to `alice_new`, then asserts only `h.balance(&alice) == Some(5_000)`. It never checks the rotation had any effect: nothing asserts `alice_new`'s key can now authorize a transfer/withdraw from that resource, and nothing asserts alice's old key can no longer.

If `apply_update_user_lock` silently dropped the write, or wrote the old lock back, this stays green, yet key rotation is an auth-transfer primitive, so the untested half is the half that matters. Suggest a follow-up action signed by `alice_new` that must succeed, plus one signed by `alice` that must be rejected. The test name promises "preserves balance **and identity**"; right now it only tests the balance.

zk/backend/risc0/test-suite/Cargo.toml
```diff
@@ -7,35 +7,47 @@ version = "0.1.0"
...
-kaspa-hashes.workspace                = true
-kaspa-rpc-core.workspace              = true
-kaspa-seq-commit.workspace            = true
-tap.workspace                         = true
-vprogs-core-codec.workspace           = true
-vprogs-core-smt.workspace             = true
-vprogs-core-types.workspace           = true
-vprogs-l1-types.workspace             = true
-vprogs-l1-wallet.workspace            = true
+k256 = { version = "0.13", default-features = false, features = [
```
hmoog: **`k256` and `signature` are pinned inline instead of going through `[workspace.dependencies]`, and they're the crypto deps.**

Lines 10 and 22 hardcode `k256 = { version = "0.13", ... }` and `signature = { version = "2", ... }`; neither is declared in the root `[workspace.dependencies]`, while every neighbour in this same file (`secp256k1`, `kaspa-txscript`, `kaspa-addresses`, all `vprogs-*`) uses `.workspace = true`.

The drift this invites is specific: test-suite's `k256` can resolve to a different patch version and feature set than the one `runtime-processor` verifies signatures with, so the harness would sign with subtly different code than the guest checks. That's the exact drift `compute_sig_message` is documented to prevent ("test harnesses sign against the same function the guest verifies (no drift)", `runtime.rs:79`). Please declare both in the root workspace and reference them with `.workspace = true`.

zk/backend/risc0/runtime-processor/src/signer_variants.rs
```diff
@@ -107,6 +108,41 @@ impl<'a> Signer<'a> for PrevTxV1WitnessSigner {
     }
 }
 
+/// Genesis-authority Schnorr signer for the bootstrap `Init` action. Unlike
```
hmoog: **The module doc's signer-kind list wasn't updated for TAG=0x05.**

Lines 5-9 enumerate "Default-build kinds, grouped by the unlocker type they produce" as 0x01 through 0x04. `GenesisSchnorrSigPtrSigner` (TAG=0x05, producing `SchnorrUnlocker`) is missing. That list is where a reviewer goes to enumerate the runtime's auth surface, and in a publicly-governed DLT an under-reported auth path is the worst kind of doc drift.

(Anchored here because the module doc itself isn't in a diff hunk.)

zk/backend/risc0/test-suite/src/runtime_flow.rs
```diff
@@ -0,0 +1,552 @@
...
+    out.push(ACTION_TAG_UPDATE_USER_LOCK);
+    out.push(user_idx);
+    new_lock.encode(&mut out);
+    out
+}
+
+// --- Assembly ---------------------------------------------------------------
+
+/// The carrier tx version the ABI decodes as `Transaction::V1`. `TX_VERSION_TOCCATA == 1`.
+const L1_TX_VERSION: u16 = 1;
```
hmoog: **Nit: `L1_TX_VERSION` hardcodes `1` with a comment asserting it equals `TX_VERSION_TOCCATA`, while the funded path imports the real constant.**

`into_unfunded()` uses this local const; the mempool path passes the genuine `TX_VERSION_TOCCATA` imported from `kaspa-consensus-core` (`runtime_flow_l1_e2e.rs:16,177`). The kaspa dep is git-branch-pinned to `master`, so the constant can move under us (likely not in practice but in theory).

If it ever does, the scheduler-driven tests keep building version-1 carriers while the mempool-driven tests use the new value, and the comment claiming they're equal becomes a lie nothing checks. Import `TX_VERSION_TOCCATA` here too.

zk/backend/risc0/test-suite/tests/runtime_flow_scenario.rs
```diff
@@ -0,0 +1,378 @@
...
+        // Below-create-minimum transfer-create -> rejected (user 8 never created).
+        TransferCreate { src: 1, dst: 8, amount: 10 },
+        // Withdraws.
+        Withdraw { user: 1, amount: 2_000 },   // u1: 2000
+        Withdraw { user: 2, amount: 50 },      // rejected: below min_withdrawal (100)
+        Withdraw { user: 3, amount: 999_999 }, // rejected: over balance
+        // Raise the withdrawal minimum, then a withdraw that the new min rejects.
+        UpdateConfig { new_min_withdrawal: 2_500 },
+        Withdraw { user: 1, amount: 2_000 }, // rejected: below new min (2500)
+        Withdraw { user: 1, amount: 2_000 }, /* accepted again? bal 2000 < 2500 min -> still
```
hmoog: **Nit: leftover uncertainty in a comment, on a duplicated action.**

`/* accepted again? bal 2000 < 2500 min -> still rejected */` reads as an unresolved question rather than a stated expectation, and the action itself duplicates line 370 exactly (same user, same amount, same rejection reason), so it adds no coverage. Either drop it or make the second one differ meaningfully. Given the "Clean up runtime flow-test comments and drop debug scaffolding" commit, I think this one just got missed.

hmoog (changes requested):
**Tracking, not blocking: this PR makes a bare signature the sole gate on `Init`, which raises the stakes on the `genesis.rs` TODO.**

`GENESIS_PUBKEY` is BIP-340 test vector 0 (private key scalar `3`, publicly known), with a TODO to replace it via a real ceremony before mainnet. Before this PR, satisfying `apply_init`'s genesis gate required a `PrevTxV1Witness`, i.e. spending a real L1 P2PK(genesis) UTXO. `GenesisSchnorrSigPtrSigner` removes that on-chain coupling: a signature alone now suffices.

Both are trivially forgeable while the key is scalar 3, so this isn't a practical regression today, and the self-contained bootstrap path is clearly the right design. But it does mean the ceremony TODO is now the *only* thing standing between the current code and "anyone can Init the config in any lane". Please make sure it's captured as a tracked issue with a release gate, not just a comment in `genesis.rs`.

No auth bypass or soundness break in the new signer.

I traced the `Init` path: `apply_init` gates on the resolved unlocker carrying exactly `GENESIS_SCHNORR_BYTES`, `compute_sig_message` length-prefixes both halves so signatures are tx-bound (no replay across the payload/rest split), and `GenesisSchnorrSigPtrSigner` grants no authority `SchnorrSigPtrSigner` doesn't already grant for live resources. Its only new capability is authorizing a resource with no readable lock, which only `Init` accepts. The harness is a genuine step up: it drives the real guest through the scheduler, the two-node L1 test proves cross-node agreement on real coinbase-funded carriers, and the settlement test asserts real journal contents.

biryukovmaxim:
Tracked the bootstrap authorization follow-up in #104. It covers both the production release gate for replacing the publicly known scalar-3 key through an agreed ceremony and the proposed refactor from a genesis-specific signer to an application-defined bootstrap lock source.

l1/wallet/src/build.rs
biryukovmaxim: Addressed in [`8b392db`](https://github.com/kaspanet/vprogs/commit/8b392db) (`Guard signed_carrier_transaction funding before the fee probe`). The extra-output value is now checked against the selected input amount before the zero-fee probe can subtract it.

zk/backend/risc0/test-suite/src/runtime_flow.rs
biryukovmaxim: Addressed in [`2f9d408`](https://github.com/kaspanet/vprogs/commit/2f9d408) (`Use assert_eq and TX_VERSION_TOCCATA in the runtime flow harness`). The scalar-3/genesis-pubkey correspondence now uses an unconditional `assert_eq!`.

zk/backend/risc0/runtime-processor/src/signer_variants.rs
biryukovmaxim: Addressed in [`9f9b19a`](https://github.com/kaspanet/vprogs/commit/9f9b19a) (`Bounds-check the genesis signer's resource_idx`) and [`9bd22e9`](https://github.com/kaspanet/vprogs/commit/9bd22e9) (`Correct the genesis signer's docs`). The signer now rejects an out-of-range target, and its docs explicitly describe the emergent Init-only restriction.

zk/backend/risc0/test-suite/tests/runtime_flow_scenario.rs
biryukovmaxim: Addressed in [`95620c2`](https://github.com/kaspanet/vprogs/commit/95620c2) (`Assert runtime scenario verdicts and exits`). The scenario now compares every journaled runtime verdict with the model, checks each emitted exit's destination and amount, and reads back the final config minimum.

zk/backend/risc0/test-suite/tests/runtime_flow_e2e.rs
biryukovmaxim: Addressed in [`1499516`](https://github.com/kaspanet/vprogs/commit/1499516) (`Verify user lock rotation transfers authority`). The test now proves the new key can withdraw from the original resource ID and that the old key is rejected without changing its balance.

zk/backend/risc0/test-suite/Cargo.toml
biryukovmaxim: Addressed in [`b798d04`](https://github.com/kaspanet/vprogs/commit/b798d04) (`Sign host-side test messages with secp256k1; keep k256 in the guest`). Rather than workspace-pinning two additional crypto dependencies, the host test harness dropped `k256` and `signature` entirely and now uses the existing workspace `secp256k1`; `k256` remains guest-only.

zk/backend/risc0/runtime-processor/src/signer_variants.rs
biryukovmaxim: Addressed in [`9bd22e9`](https://github.com/kaspanet/vprogs/commit/9bd22e9) (`Correct the genesis signer's docs`). The module-level signer-kind list now includes TAG `0x05`.

zk/backend/risc0/test-suite/src/runtime_flow.rs
biryukovmaxim: Addressed in [`2f9d408`](https://github.com/kaspanet/vprogs/commit/2f9d408) (`Use assert_eq and TX_VERSION_TOCCATA in the runtime flow harness`). The unfunded carrier path now imports and uses `TX_VERSION_TOCCATA` directly.

zk/backend/risc0/test-suite/tests/runtime_flow_scenario.rs
biryukovmaxim: Addressed in [`98fe4e3`](https://github.com/kaspanet/vprogs/commit/98fe4e3) (`Drop the duplicate withdraw in the deterministic scenario`). The duplicate action and uncertain comment were removed.

hmoog (approved):
Looks good now

PR #96, opened 2026-06-30
Runtime: L1-backed deposits, single creation policy, within-tx double-deposit fix

### LINKED ISSUE #92: Within-tx resource lifecycle uses static is_new flag, not a state machine — same-tx double-deposit silently overwrites (fund loss)
## Summary

A resource's lifecycle within a single transaction is inferred from `Resource::is_new()` — a **static input flag** set once at decode from prior-state existence — plus current data-emptiness. Nothing advances this lifecycle as actions apply *within* the same tx. When two actions in one tx touch the same resource, the second reads a stale lifecycle and can silently clobber the first.

Concretely for deposits: two `Deposit` actions to the same brand-new user in one tx do **not** accumulate (create then credit). The second deposit overwrites the first.

## Reproduction

`deposit_create_then_credit_same_user` (`#[ignore]`d) in `zk/backend/risc0/runtime-processor/tests/e2e.rs` — two `Deposit` actions to one new user (`v0=5000`, `v1=3333`). Expected `8333`, observed `3333` (the second overwrites the first).

## Root cause

`Resource::is_new()` returns a field set once at decode (`zk/abi/src/transaction_processor/input/resource.rs`):

```rust
pub fn is_new(&self) -> bool { self.is_new }
```

`resize()` and `data_mut()` only set `dirty = true`; they never clear `is_new`. So both same-tx actions observe `is_new() == true`, both take the create branch, and the second `init_user` — guarded only by `if !self.is_new()` (`runtime-processor/src/resource_ext.rs`) — proceeds and overwrites the slot written by the first.

The create-or-credit "decided from live slot state" property therefore holds only **across** txs (where the updated state root flips `is_new` for the next tx), **not within** one tx.

## Impact

Fund-loss path: `v0`'s funding output is consumed and its value enters the covenant on L1, but the resulting L2 balance reflects only `v1`. The first deposit's credit is dropped. This is distinct from the cross-batch anti-double-count follow-up noted in `deposit_policy.rs` — that guards replay; this is a within-tx lifecycle-staleness bug.

## Ephemeral flows (new -> live -> deleted within one tx)

The deeper problem the state machine needs to support is a resource that is created, used, and torn down entirely inside a single tx — never touching committed state. Example: a scratch / aggregator account that one action creates (`New -> Live`), a second action drains or finalizes, and a third action deletes (`Live -> Deleted`), all in the same tx.

This is **not reachable at all today**, for two compounding reasons:

1. **No delete action exists.** Nothing in the runtime calls `resize(0)` / empties a resource, so `Live -> Deleted` cannot be driven through the action API. Deletion is only ever inferred from `data().is_empty()` (`resource_ext.rs` treats `is_new() || data().is_empty()` as "not live").

2. **"Deleted" is conflated with "new/never-existed", and `is_new` is frozen.** Because deletion is just empty data and `is_new` never advances, a torn-down slot is indistinguishable from a never-created one. A later same-tx action would either silently re-create it (the `init_*` guard is only `!is_new()`, still `true`) or mis-read it as empty — there is no `Deleted` terminal state to reject `use-after-delete` or `re-create-after-delete`.

Even the prerequisite `New -> Live` transition is invisible intra-tx — in `deposit_create_then_credit_same_user` the second deposit takes the create branch even though the slot was just made live — so none of the ephemeral lifecycle invariants can be enforced.

## Proposed fix

Replace the static `is_new` snapshot with a proper per-resource lifecycle state machine that the runtime advances as actions apply within a tx — e.g. `New -> Live -> Deleted`, with create/init/credit/delete transitions checked against the **current** state, not the input snapshot, and a real `Deleted` terminal state distinct from `New`. Apply fns (`init_user`, `apply_deposit`, deletion paths) would then branch on live lifecycle state, so:

- a second same-tx action correctly sees the slot as `Live` (credit) instead of re-creating it;
- ephemeral `New -> Live -> Deleted` is expressible, with `use-after-delete` and `re-create-after-delete` rejected;
- illegal transitions (double-create) error rather than silently overwriting.

Un-ignore `deposit_create_then_credit_same_user` once fixed (and add a `New -> Live -> Deleted` test once a delete action lands).

biryukovmaxim:
> Stacks on **runtime-poc** (expected to land first). The description below covers
> only the work added on top of that branch; the diff will collapse to those
> commits once runtime-poc merges.

## Summary

Hardens the runtime's user-creation and deposit model and fixes a within-tx
double-deposit bug, plus a small signer-hashing optimization and a no-op module
split.

## L1-backed deposits & one creation policy

- **Bind L2 deposits to a per-covenant on-chain address.** Deposit funding
  outputs must pay `P2SH(delegate_entry_script(covenant_id))`, so credited L2
  balance is always backed by a covenant-spendable L1 UTXO. The deposit
  script-hash (`deposit_spk_hash`) is threaded through per-tx exits, the batch
  and aggregator journals, the output commitment, and the verifiers.
- **Make Deposit the sole user-creation path.** Removed the auth-free,
  zero-balance `UserInit` action (and renumbered the remaining action tags), so
  a user is born only when funded. A policy-enforced `min_create_balance`
  forbids zero-balance accounts.
- **Let Transfer create its destination** under the *same* gate as Deposit —
  address binding (slot id derives from the supplied initial lock) plus the
  creation minimum — via a shared `validate_user_create`. `Transfer` gains an
  optional dest-lock (one presence byte: create-if-new vs. credit-existing).
  Still no unbacked, zero-balance birth.

## Fix within-tx double-deposit (#92)

The create-vs-credit decision read `Resource::is_new()`, a static per-tx flag
that never advanced as actions applied — so two deposits to one new user in the
same tx both took the create branch and the second overwrote the first (the L1
funding entered the covenant but only the second balance survived).

Now the runtime's `ApplyContext` tracks per-resource lifecycle
(`New -> Live -> Deleted`), seeded from each decoded snapshot and advanced as
actions apply. Apply fns decide create-vs-credit from live state and drive
`New -> Live` via `mark_created`, which rejects double-create and
re-create-after-delete. The ABI `Resource` stays lifecycle-agnostic; the view
layer reads liveness off data-emptiness. Un-ignores
`deposit_create_then_credit_same_user`.

## Misc

- **Lazy schnorr sig-message hashing.** The sig-message digest is now computed
  on first use behind a `OnceCell` in `SignerResolveContext`; witness-only /
  signer-less txs no longer pay a wasted SHA-256.
- **Split `runtime-processor/action.rs`** into per-action modules
  (`config` / `user` / `withdraw` / `deposit`); `action.rs` keeps `ApplyContext`
  and dispatch. No behavior change.
- Guest ELFs rebuilt.

## Tests

New/updated e2e and unit coverage: deposit create-gate (below-minimum,
address-binding), transfer create / below-minimum / missing-lock, lifecycle
transitions, and the un-ignored within-tx create-then-credit repro.

## Closes

Closes #92


zk/abi/src/batch_processor/verifier.rs
```diff
@@ -209,3 +216,20 @@ impl<'a, V: FnMut(&[u8; 32], &[u8])> Verifier<'a, V> {
...
+/// hash in it is the same value; a second differing non-zero hash is an inconsistency the release
+/// path carries opaquely and a debug build rejects.
+fn carry_deposit_hash(carry: &mut [u8; 32], tx_hash: &[u8; 32]) {
+    if *tx_hash == [0u8; 32] {
+        return;
+    }
+    if *carry == [0u8; 32] {
+        *carry = *tx_hash;
+    } else {
+        debug_assert_eq!(carry, tx_hash, "two distinct deposit_spk_hash values in one batch");
```
hmoog: **`debug_assert_eq!` is compiled out in the proving guest, so this conflict check does nothing in release.**

The guest builds in release, where `debug_assert_eq!` is a no-op. On two differing non-zero `deposit_spk_hash` values in one batch, this silently keeps the first and drops the rest instead of rejecting. That's consensus-critical: the second deposit's on-chain funding output would never get bound by settlement.

It's also asymmetric with the per-tx path: `DepositSink::set` (`deposit_sink.rs`) returns a hard `Err` for the exact same conflict. The safety argument ("a bundle settles into one covenant, so all hashes are identical") rests on an *unasserted* "one config/covenant per batch" invariant; nothing here cross-checks each tx's deposit-deriving covenant against the batch `covenant_id`.

Please make this a hard `assert_eq!` (or return an error) to match `DepositSink::set`.

zk/abi/src/batch_aggregator/verifier.rs
```diff
@@ -160,3 +167,20 @@ where
...
+/// hash in it is the same value; a second differing non-zero hash is an inconsistency the release
+/// path carries opaquely and a debug build rejects.
+fn carry_deposit_hash(carry: &mut [u8; 32], batch_hash: &[u8; 32]) {
+    if *batch_hash == [0u8; 32] {
+        return;
+    }
+    if *carry == [0u8; 32] {
+        *carry = *batch_hash;
+    } else {
+        debug_assert_eq!(carry, batch_hash, "two distinct deposit_spk_hash values in one bundle");
```
hmoog: **Same issue as the batch-processor `carry_deposit_hash`: `debug_assert_eq!` is a no-op in the guest.**

In release this silently keeps the first non-zero `deposit_spk_hash` and drops a differing one instead of rejecting. Should be a hard `assert_eq!` / error, consistent with the per-tx `DepositSink::set` behaviour. (See the matching comment on `batch_processor/verifier.rs:233`.)

zk/backend/risc0/runtime-processor/src/tx_inputs.rs
```diff
@@ -71,10 +71,10 @@ pub struct OutputData<'a> {
...
-pub fn parse_output_at_index_v1<'a>(
-    rest_preimage: &'a [u8],
+pub fn parse_output_at_index_v1(
+    rest_preimage: &[u8],
     output_index: u32,
-) -> CodecResult<OutputData<'a>> {
+) -> CodecResult<OutputData<'_>> {
     let mut buf = rest_preimage;
     let version = buf.le_u16("tx.version")?;
     if version < 1 {
```
hmoog: **Version guard is looser than the documented contract: accepts v2+ and parses them with the V1 layout.**

The doc says "only `tx_version == 1` is supported (V1 outputs include the `has_covenant` byte and optional covenant tail)", but `if version < 1` only rejects version 0. Any version ≥ 2 is parsed as V1: reading a `has_covenant` byte + covenant tail a future layout may not have, silently misaligning subsequent output boundaries.

This runs on the attacker-supplied `witness_rest_preimage` in `recover_prev_tx_v1_p2pk_pubkey` (auth path). The preimage is pinned to a real L1 tx-id, but the extractor's soundness assumes the V1 output layout, so a ≥2 version tx would misparse rather than reject.

Suggest `if version != 1` to match the stated contract.

zk/backend/risc0/runtime-processor/src/lock_codec.rs
hmoog, line 46: **`pub` decoder indexes untrusted bytes (`body[0]`, `&body[2..]`) with no length guard: panics unless the caller ran `validate_lock_body` first.**

The Multisig arm assumes `body.len() >= 2`; it's only sound because every current caller (`ConfigView`/`UserView::from_bytes`) validates first and then `.expect(...)`. A `pub` function that indexes untrusted input on an unenforced precondition is a footgun: any future caller that forgets the pairing gets an OOB panic, which in the guest means an un-provable tx (DoS).

Suggest making this `pub(crate)` and/or re-checking length defensively. (The threshold/length/ordering logic is also duplicated with `MultisigLockView::decode`; worth a shared helper so they can't drift.)

zk/backend/risc0/runtime-processor/src/ix.rs
hmoog: **Nit / consistency: signer `resource_idx` is only ordering-checked here, not bounds-checked against `n_resources`.**

Action indices all go through `read_resource_idx` (bounds-checked), but signer `resource_idx` only gets the ascending-order check. It's fail-closed today (`resources.get(idx)` errors for signature signers; witness signers ignore the idx), so this isn't a bypass: just flagging the asymmetry; a bounds check here would make the two paths consistent.

zk/backend/risc0/api/src/witness.rs
```diff
@@ -50,7 +55,14 @@ impl OwnedSuccinctWitness {
         let control_digests =
             s.control_inclusion_proof.digests.iter().flat_map(|d| <[u8; 32]>::from(*d)).collect();
 
-        Self { seal, claim, control_index, control_digests }
+        Self { seal, claim, control_index, control_digests, deposit_spk_hash: [0u8; 32] }
```
hmoog: **`deposit_spk_hash` defaults to `[0u8; 32]`: a caller that forgets `with_deposit_spk_hash` on a deposit-crediting bundle produces an unprovable settlement.**

This is a liveness footgun, not a security hole: a forgotten setter yields a journal preimage that won't match the guest's committed hash, so the proof fails. Only the succinct path has a caller today; `OwnedGroth16Witness` (line ~109) has the same default and no caller yet. If a Groth16 settlement builder is added it must not forget it. Consider a debug-assert at build time, or making the field non-defaulted.

zk/backend/risc0/covenant/src/script.rs
hmoog, line 912: **`perm_value` / permission-output value is baked into the redeem-script bytes with no guard tying it to the P2SH continuation chain.**

Because this value is a script constant, it participates in the P2SH hash that chains one covenant UTXO to the next. It's constant everywhere today (`DEFAULT_PERMISSION_OUTPUT_VALUE`), so this is fine but if it's ever made per-settlement, `next_redeem` must be rebuilt with the *identical* value or the chain silently breaks (the next spend rebuilds a different script and fails the SPK check). Worth a comment or assert documenting that coupling.

Not something that should / has to be fixed - Just flagging here because I am not sure how the remaining upstream PR's will handle this.

zk/backend/risc0/runtime-processor/src/runtime.rs
hmoog: **Minor: the sig-message hashes `domain || rest_preimage || payload_presig` with no length delimiter between the two variable-length halves.**

Two different `(rest_preimage, payload_presig)` splits with the same concatenation would collide. Not practically exploitable (both halves are independently structurally constrained, so a shifted split won't reparse validly), but a length prefix on the first half would remove the ambiguity class entirely.

Up to you to decide if this should be handled or we rely on the structural constraints.

zk/backend/risc0/runtime-processor/src/deposit_policy.rs
```diff
@@ -0,0 +1,199 @@
...
+//!
+//! Deposit-address binding: a deposit's funding output must pay
+//! `P2SH(delegate_entry_script(config.covenant_id))`, the same covenant-spendable
+//! delegate-entry script the permission-script sweep recognises. `covenant_id`
+//! is sourced from the config resource (committed L2 state, set once at `Init`)
+//! rather than baked into the guest, so the zkVM image id is invariant to it;
+//! the binding lives in committed state (its bytes hash into the resource
+//! commitment and the SMT state root).
+//!
+//! Follow-ups (not in this unit):
```
hmoog: **Tracking the two documented follow-ups: Maybe they should be captured as issues before this leaves POC status:**

1. **Batch-tier covenant assertion** (`config.covenant_id == settlement covenant_id`): today the on-chain `OpEqualVerify` binds them, so a mismatch is stuck-funds (liveness) rather than theft: but the in-circuit assert is the missing belt to that suspenders.
2. **Cross-batch deposit anti-double-count**: within a tx `consumed_outputs` dedups, and funding is an output of the current tx (so cross-tx uniqueness comes from L1 tx uniqueness). The one residual inflation path is the *same tx admitted into two batches*: please confirm lane/sequencing rules this out, since that's the difference between "documented POC gap" and "double-mint."

Note: I think 2. it is enforced by the accepted txs being deduplicated across merge-sets on the L1 but just flagging it anyway for soundness (can most probably be ignored).

hmoog (changes requested):
Looks good with a few minor issues.

Feel free to respond / ignore the ones that you think are not as necessary / pressing for the MVP or that are being addressed by further pending PR's.

zk/abi/src/batch_processor/verifier.rs
biryukovmaxim: Fixed in 693fa52

zk/abi/src/batch_aggregator/verifier.rs
biryukovmaxim: Fixed in the same commit, 693fa52 - promoted to assert_eq! alongside the batch-processor one.

zk/backend/risc0/runtime-processor/src/tx_inputs.rs
biryukovmaxim: Fixed in 693fa52

zk/backend/risc0/runtime-processor/src/lock_codec.rs
biryukovmaxim: Split in 693fa52: `decode_lock_body` is the pub entry point, validates first and returns Result; `decode_lock_body_unchecked` is now pub(crate) and states the precondition

zk/backend/risc0/runtime-processor/src/ix.rs
biryukovmaxim: Fixed in https://github.com/kaspanet/vprogs/commit/693fa52bda1c0a62bb958f62f8846c87850b42cb. decode_ix now bounds-checks signer resource_idx against n_resources

zk/backend/risc0/api/src/witness.rs
biryukovmaxim: Fixed in 693fa52

zk/backend/risc0/covenant/src/script.rs
biryukovmaxim: Documented in 693fa52

zk/backend/risc0/runtime-processor/src/runtime.rs
biryukovmaxim: Fixed in 693fa52. Both halves are length-prefixed with LE u32: `SHA-256(domain || len(rest_preimage) || rest_preimage || len(payload_presig) || payload_presig)`

zk/backend/risc0/runtime-processor/src/lock_codec.rs
biryukovmaxim: duplication is gone as of abdf278

hmoog (approved):
Looks good - let's merge

PR #97, opened 2026-07-01
vprogs-runner: program-agnostic daemon engine + vprun CLI, and a live account-model example

biryukovmaxim:
## Stacked on the runtime POC

This branch is **stacked on top of [#96](https://github.com/kaspanet/vprogs/pull/96) (`runtime-poc-continue`)**, which is itself stacked on [#42](https://github.com/kaspanet/vprogs/pull/42) (`runtime-poc`). Until those merge, the GitHub diff below includes their commits too. **This description covers only the changes on top of `runtime-poc-continue`** (i.e. `runtime-poc-continue..HEAD`); review the base PRs separately.

## What this adds on top of the runtime POC

The runtime POC proved out the guest-side account model and settlement. This stack turns the tn10-flow proof-of-concept binary into a reusable **daemon engine + CLI**, and adds a second example that drives the account-model guest end to end against a live testnet-10 fork node.

### `vprogs-runner` crate + `vprun` CLI

A program-agnostic daemon engine extracted from the tn10-flow POC: it follows a Kaspa L1 lane, executes an arbitrary guest program over its transactions, and optionally proves + settles the results. It only fetches/executes/proves/settles; **issuing lane transactions is left to the caller** (the examples).

- Two run modes: execution-only, or the full transaction/batch/aggregate proving stack with an on-chain settlement worker.
- Three explicit start modes (`fresh` / `resume` / `catchup`), replacing tn10-flow's implicit inference, with the safety guards preserved as typed errors (catch-up needs a deploy block; resume needs persisted state).
- `vprun` CLI with layered config, highest precedence first: CLI flags → `VPRUN_*` env → `--config` TOML → defaults.
- A small JSON state file persists lane id, covenant id, bootstrap txid, and the deploy-block anchor so a restart resumes forward without re-supplying them.

### `l1/wallet` and `l1/utils` crates

- `l1/wallet`: keypair-backed L1 transaction building (covenant bootstrap, lane activity, settlement funding, deposits, pay-to-address) over any `RpcApi` client, with no consensus internals, factored out of the POC so it works in-process or over RPC.
- `l1/utils`: byte-oriented v1 `tx_id` / `payload_digest` helpers usable from the zkVM guest.

### `tn10-runtime` example

A new driver that exercises the `runtime-processor` account model (Init / Deposit / Transfer / Withdraw) against a remote tn10 fork node:

- Genesis `Init` authorized by an **L1 prev-tx witness** (spending a `P2PK(GENESIS)` output) rather than a hand-seeded config, so the guest recovers the genesis pubkey from the spent output.
- Deposit fee **iterated to a fixpoint** so the built tx clears the mempool floor recomputed on its real (shrunk) change output under KIP-9 storage mass.
- 2-node demo scripts + a **follower mode** (`TN10RT_ISSUE=0`) so multiple nodes can settle one covenant without duplicating the once-only Init.

### tn10-flow refactor

`tn10-flow` is now a thin driver over the runner: it keeps only the background activity issuer and delegates fetch/execute/prove/settle to the engine. The `two_provers_contend` test is ported onto the new crate.

### Resilience

Transient wRPC failures against a live node no longer take the daemon down:

- Bounded retries with backoff for the lane-proof fetch, settlement funding, and bridge-seed anchor resolution.
- The activity issuer logs and retries on a spendable-set fetch error instead of crashing.
- Fixed a bridge reorg-past-root panic by seeding the fresh-chain root with real reorg headroom (pin an anchor only when it is already deep, else seed `seed_depth` below the sink).
- Fixed the live no-settlement stall this seed change exposed: the batch prover proved every empty (no-lane-tx) backfill block in order, so a node seeded deep below the deploy block never reached the first lane block to form a bundle. It now skips proving empty batches (their receipts are never composed by the aggregate prover, which filters on the same `txs().is_empty()` predicate), so settlement fires promptly regardless of seed depth and per-block proving cost no longer scales with backfill length (essential under real CUDA proving). Validated live against tn10 with confirmed on-chain settlements, and by the dev-mode settlement suite.

## Testing

- `deposit.rs` unit tests cover the fixpoint fee against the node floor on shrunk change.
- `runner/src/config.rs` covers config layering precedence and validation.
- `tn10-runtime/tests/runtime_actions.rs` is a direct-guest acceptance test over the same encoders the driver ships.
- The account-model live sequencing (the driver's step pacing) is POC-only and not covered by tests; this is called out in the module docs.


runner/src/start.rs
```diff
@@ -0,0 +1,625 @@
...
+                return Err(StartError::DataDirNotClean(cfg.data_dir.clone()));
+            }
+            // Capture the node's selected tip before bootstrap: a real chain block at or just
+            // before the deploy block, seeded so a later resume replays forward from
+            // the deploy.
+            let seed_block = client.get_block_dag_info().await.expect("get_block_dag_info").sink;
+            let wallet = Wallet::new(client, params, keypair);
+            log::info!("bootstrapping dev covenant; issuer address {}", wallet.address());
+            let (covenant, bootstrap_txid) =
+                bootstrap_dev_covenant(&wallet, lane_key, COVENANT_VALUE).await;
```
hmoog: Exec-mode `Fresh` always mints a **dev** covenant, while `start_settlement` branches on `dev_mode_enabled()` (lines 290 and 373).

The footgun: a fresh exec run persists that dev-pins covenant id, and a later restart of the same data dir with `--prove` in non-dev mode reconstructs the redeem with real pins (line 337), producing an SPK that cannot match the on-chain dev UTXO. Nothing fails until the first settlement's SPK assert, far from the cause. Either branch on `dev_mode_enabled()` here the way settlement mode does, or reject non-dev exec-fresh with a typed `StartError` naming the constraint.

runner/src/main.rs
```diff
@@ -0,0 +1,146 @@
...
+    /// var.
+    #[arg(long)]
+    config: Option<std::path::PathBuf>,
+
+    /// Borsh wRPC URL of the node, e.g. `ws://host:17210`.
+    #[arg(long)]
+    wrpc_url: Option<String>,
+    /// Fee / bootstrap secp256k1 secret key (32-byte hex).
+    #[arg(long)]
+    private_key: Option<String>,
```
hmoog, line 29: `--private-key` as a CLI flag leaks the key into shell history and the process list (`ps`). The env and TOML layers already exist; recommend the docs (and the flag's help text) steer operators to `VPRUN_PRIVATE_KEY` or the config file, keeping the flag as a last resort for throwaway dev keys.

hmoog (changes requested):
Looks good - just two minor things.

PR #102, opened 2026-07-13
Fix reorg settlement stalls: wake parked waiters on rollback cancellation (#98-#101)

### LINKED ISSUE #98: Settlement permanently stalls after a reorg: rollback cancellation never wakes parked latch waiters
**Severity: production liveness. Silent, unbounded, no self-heal; only a restart recovers.**

A rollback cancels in-flight batches by lowering an `AtomicU64` threshold (`scheduling/scheduler/src/cancellation_context.rs`), which wakes nothing. `ScheduledBatch::wait_artifact_published` samples `canceled()` once, then parks on the latch (`scheduling/scheduler/src/scheduled_batch.rs`). For a non-empty batch that finished executing while live, the only cancel-side opener (the canceled branch at the end of `decrease_pending_txs`) has already declined, and the batch prover skips publishing for a canceled batch (`zk/batch-prover/src/worker.rs`), so the latch never opens and a waiter that parked before the rollback is stuck forever.

The aggregate prover parks exactly there with no rollback escape (`zk/aggregate-prover/src/worker.rs`, `try_prove_one_bundle`: a `select!` over shutdown + the latch only). Parked, it can never drain the `Command::Rollback` that would evict the canceled front. Bundling stops, the settlement queue drains, the settler parks. The node keeps following the chain and executing, logs nothing, and never settles again.

Observed live on tn10 with real proofs: 6 settlements in the first 90 seconds, then zero for the remaining 9+ minutes while lane txs, execution, and the L1 bridge all stayed healthy. It is a race window, so in production it surfaces intermittently as "settlements mysteriously stopped".

### Reproduction (deterministic, no GPU, no network, ~5s)

Branch [`scheduler-stall-repros`](https://github.com/biryukovmaxim/vprogs/tree/scheduler-stall-repros) (3 commits off master, one per related issue; the failing tests are `#[ignore]`d so the branch is CI-green — un-ignoring a test is the acceptance criterion for its fix):

```
cargo test -p vprogs-scheduling-scheduler --test e2e test_wait_artifact_published_wakes_on_late_cancellation -- --ignored --exact
```

Fails at the `recv_timeout` assertion: a consumer parked on the canceled batch's artifact latch is never released.

### Fix direction

Cancellation must *wake* parked waiters, not merely be pre-checked — e.g. a per-batch canceled latch registered on the cancellation context, opened when the threshold walk cancels the batch, with every `wait_*` becoming a two-arm select. Whatever the shape, it must cover the whole `wait_*` family (see the companion issues) or the stall just moves one stage earlier.

### LINKED ISSUE #99: wait_committed: cancellation before commit scheduling strands waiters and hangs daemon teardown
**Severity: liveness + shutdown deadlock.** Same wake gap as #98, on the `committed` latch — arguably worse.

`committed` has no cancel-side opener at all: it opens only in `commit_done`, reached only through `schedule_commit`, which is a no-op for a canceled batch (`scheduling/scheduler/src/scheduled_batch.rs`). A batch canceled before the lifecycle worker schedules its commit never opens the latch; a waiter that parked earlier is stuck forever.

The batch prover parks on `wait_committed` for every batch — including the empty batch produced for every chain block without lane transactions, i.e. the overwhelming majority of blocks — with no escape arm (`zk/batch-prover/src/worker.rs`). This both stalls proving (which starves the aggregate prover and the settler, same silent cascade as #98 one stage earlier) and hangs teardown: `Scheduler::shutdown` stops the batch lifecycle worker before joining the provers, so the batch prover's `join()` never returns and the process cannot exit. A reorg racing a Ctrl-C wedges the daemon.

### Reproduction (deterministic, ~5s)

Branch [`scheduler-stall-repros`](https://github.com/biryukovmaxim/vprogs/tree/scheduler-stall-repros):

```
cargo test -p vprogs-scheduling-scheduler --test e2e test_wait_committed_wakes_on_late_cancellation -- --ignored --exact
```

The test gates execution of the preceding batch with a parked transaction, so the FIFO lifecycle worker provably has not scheduled the watched batch's commit when the rollback lands; the parked waiter is never released.

### Fix direction

Covered by the same cancellation-wake mechanism as #98 — it must include `committed`. The batch prover's parks should additionally be shutdown-escapable so a wedged latch can never hang teardown.

### LINKED ISSUE #100: wait_tx_artifacts_published: missing cancellation wake (masked contract gap)
**Severity: latent; contract-level.** Same wake gap as #98, on the `tx_artifacts_published` latch — currently masked by publisher behavior.

A waiter that parked on `wait_tx_artifacts_published` while the batch was live is not released when a rollback later cancels the batch. The unconditional opener (`decrease_pending_tx_artifacts`) fires only once every transaction publishes its artifact; the cancel-conditional opener in `decrease_pending_txs` fires only if the batch was already canceled when its last transaction finished. Neither covers a cancellation landing after the waiter parked when nothing publishes (`scheduling/scheduler/src/scheduled_batch.rs`).

In production the transaction prover currently publishes `None` for every transaction of a canceled batch (`zk/transaction-prover/src/worker.rs`), so this latch does drain today. But the waiter is the batch prover (`zk/batch-prover/src/worker.rs`), the latch contract is broken, and it is one publisher change away from parking the batch prover forever — with the same teardown-hang consequence as #99, since that park has no escape arm either.

### Reproduction (deterministic, ~5s)

Branch [`scheduler-stall-repros`](https://github.com/biryukovmaxim/vprogs/tree/scheduler-stall-repros):

```
cargo test -p vprogs-scheduling-scheduler --test e2e test_wait_tx_artifacts_published_wakes_on_late_cancellation -- --ignored --exact
```

(The harness has no transaction prover, so nothing publishes and only the missing cancellation wake could release the waiter — which is exactly the contract the fix must provide.)

### Fix direction

Covered by the same cancellation-wake mechanism as #98 — apply it uniformly to the whole `wait_*` family rather than per-latch.

### LINKED ISSUE #101: Aggregate prover panics when a bundle sweeps in a batch canceled mid-execution
**Severity: crash — the aggregate worker thread dies, and bundling/settlement stop until restart.**

Found by code review during the #98 investigation. Same reorg-racing-execution window as #98, different failure mode (panic instead of park).

The bundle extend loop (`zk/aggregate-prover/src/worker.rs`, `try_prove_one_bundle`) includes any queued batch whose `artifact_published()` latch is open:

```rust
while take < self.queued.len()
    && take < *self.bundle_size.end()
    && self.queued[take].artifact_published()
```

But `decrease_pending_txs` force-opens that latch *without storing a receipt* for a batch whose last transaction finished after cancellation (`scheduling/scheduler/src/scheduled_batch.rs`). A canceled batch sitting behind a still-live, published front therefore gets swept into a bundle, and `prove_bundle`'s receipt collection

```rust
.filter(|b| !b.txs().is_empty())
.map(|b| (*b.artifact()).clone())
```

calls `ScheduledBatch::artifact()`, which panics with `"batch artifact not ready"`. The filter excludes empty batches, not canceled ones.

Reachable interleaving: batches [A live, B executing] queued; a rollback cancels B (threshold below B's index) while A is still being proved; B's last transaction finishes after the cancellation, so the canceled branch force-opens B's artifact latch; A's receipt publishes; the extend loop takes A and B; `prove_bundle` panics on B.

### Deterministic repro (no GPU, no network, ~5s)

Branch `scheduler-stall-repros`, `zk/aggregate-prover/tests/canceled_batch_bundle.rs`:

```
cargo test -p vprogs-zk-aggregate-prover --test canceled_batch_bundle -- --ignored
```

The test drives the real `AggregateProver` worker over a mock backend whose proving methods are all `unreachable!()` (every bundle it forms is all-empty, so nothing is ever proved). A gate processor holds the canceled batch's only transaction inside execution, so the rollback provably lands while it is still in flight; releasing the gate then force-opens its artifact latch with no receipt behind it. The front is an empty batch, which is live and published from creation, as every chain block without lane transactions is.

The rollback reaches the scheduler but not the prover's inbox, matching production: the worker is parked on the front's artifact latch when the reorg lands and cannot drain `Command::Rollback` until it returns from bundling, which is the very call that panics.

Fails 10/10 runs today:

```
thread '<unnamed>' panicked at scheduling/scheduler/src/scheduled_batch.rs:189:35:
batch artifact not ready
```

The test also probes worker liveness with a following batch, so it fails whether the canceled batch is swept in behind the front (extend loop) or reached as the front itself after the front bundles alone — a worker that panicked collecting its receipt never bundles again. `#[ignore]`d so CI stays green; un-ignoring it is the acceptance criterion for the fix.

### Fix direction

Stop the extension at the first canceled batch (`&& !self.queued[take].canceled()` before the latch check). `canceled()` is monotonic and a live non-empty batch with an open latch always has a real receipt, so the check is race-free. A canceled batch reached as the front must also be evicted rather than bundled; evicting the rest of the canceled suffix remains the rollback command's job.

biryukovmaxim:
A reorg lowers a `CancellationContext`'s threshold to cancel the batches it rolls back. The scheduler's `wait_*` methods sampled `canceled()` once and then parked on a latch the threshold store never touched, so any waiter already parked when the rollback landed stalled forever. On live tn10 this wedged settlement permanently after a reorg.

## Root cause

Check-then-park with no wake on cancellation:

- `wait_artifact_published`, `wait_tx_artifacts_published`, and `wait_committed` parked on latches a rollback never reopened (#98, #99, #100).
- A batch canceled mid-execution had its `artifact_published` latch force-opened with no receipt; the aggregate-prover's extend loop swept it into a bundle and panicked collecting the missing receipt (#101).

## Fix

- **`WaitCell`** (`core/atomics`): an `AtomicU64` paired with a tokio `Notify` so every store wakes all parked waiters, with a level-triggered `wait_until(pred)` and a blocking twin. Replaces the now-unused `AtomicEnum`.
- **scheduler**: hold the cancellation threshold in a `WaitCell`. Each `wait_*` races its latch against `wait_canceled(index)`, which resolves once the threshold drops below the batch index, releasing a batch canceled while its waiter is parked.
- **aggregate-prover**: evict a canceled front instead of bundling it, and check `artifact_published()` before `canceled()` in the extend loop so a receiptless force-opened latch never extends the bundle.
- **batch-prover / settler**: make the tx-artifacts and commit parks shutdown-escapable, and warn every 30s while a settlement sits unconfirmed.

## Tests

Each issue ships a deterministic repro (the first four commits), un-ignored by its fix as the acceptance criterion:

- scheduler e2e wake repros + a deep-rollback re-park test — **39 passed**
- aggregate-prover sweep repro + whole-queue eviction — **2 passed**
- WaitCell waker-contract suite — **12 passed**

Fixes #98
Fixes #99
Fixes #100
Fixes #101


hmoog (approved):
It's clean - let's merge!

PR #117, opened 2026-07-17
Bind the resource lifecycle bit by deriving it from resource data

### LINKED ISSUE #105: Runtime trusts unbound is_new resource flag
## Summary

The runtime processor trusts the per-resource `is_new` bit from transaction input even though that bit is not committed in the transaction journal and is not checked by the batch verifier.

For an existing non-empty resource, a prover can keep the real resource bytes, so the resource hash still matches the SMT value, while changing only `is_new=true`. The runtime lifecycle code then classifies the resource as `New`.

## Impact

This can change guest control flow for creation-sensitive actions. The repro demonstrates the config `Init` path accepting an existing non-empty config as new when only `is_new` is forged.

The old-root/resource-hash check does not catch this because it binds the resource bytes and id, not the lifecycle bit. In the repro the live config bytes are genuine, so the old SMT value remains derivable.

## Reproduction

Repro branch:

https://github.com/biryukovmaxim/vprogs/tree/repro/g12-is-new-unbound

Commit:

`fc96307 Reproduce forged is_new lifecycle input`

Focused command:

```bash
RISC0_DEV_MODE=1 cargo test \
  --manifest-path zk/backend/risc0/runtime-processor/Cargo.toml \
  --test e2e \
  live_config_falsely_marked_new_is_rejected \
  -- --nocapture
```

```text
test live_config_falsely_marked_new_is_rejected ... FAILED

thread 'live_config_falsely_marked_new_is_rejected' (...) panicked at tests/e2e.rs:824:5:
guest accepted Init over a live config when only is_new was forged
```

## Root Cause

`Resource::decode` reads `is_new`, `index`, and `data` from host/prover-supplied input.

`InputResourceCommitment` commits:

- resource index;
- resource id;
- hash of resource data.

It does not commit `is_new`.

The batch verifier checks the committed data hash and resource id against the SMT proof, but it has no visibility into the lifecycle bit. Runtime lifecycle classification currently treats `(is_new=true, data=anything)` as `New`.

## Suggested Fix

Enforce the honest invariant at the ABI boundary:

```rust
is_new => data.is_empty()
```

A live non-empty resource must not decode as new. The likely enforcement point is `zk/abi/src/transaction_processor/input/resource.rs` in `Resource::decode`, after reading `is_new`, `index`, and `data`.

Suggested regression coverage:

- ABI unit test: new resource with empty data is accepted.
- ABI unit test: live resource with non-empty data is accepted.
- ABI unit test: new resource with non-empty data is rejected.
- Runtime focused test: `live_config_falsely_marked_new_is_rejected` passes after rebuilding the runtime guest ELF.

## Related Hardening Notes

Adjacent fields at the same boundary appear to be covered differently:

- `context_hash` is journaled and compared against batch metadata-derived context.
- `tx_id`, `version`, and `merge_idx` are journaled and consumed by the batch verifier/activity logic.
- resource id and data are journaled and checked against the SMT proof.
- resource index is journaled and checked, but depends on proof canonicality. That dependency is now its own issue: **#106** (batch verifier does not canonicalize prover-supplied proof structure), which stands up the `batch_processor::Verifier` harness this issue lacked. Neither blocks the other.
- `access_type` is committed through `tx_id`, but its enforcement is a separate host/guest consistency issue.

So this issue is specifically the unbound lifecycle bit, not a failure of the existing old-root derivation.

biryukovmaxim:
Fixes #105.

## Problem

The transaction input wire carried a per-resource `is_new` flag that nothing committed. `InputResourceCommitment` binds only the resource index, id, and data hash, and the batch verifier checks those against the SMT proof, so it had no visibility into the lifecycle bit.

A prover could keep an existing resource's genuine bytes (leaving the old-root check satisfied) and flip `is_new=true`, making `Lifecycle::from_resource` classify a live slot as `New`. The config `Init` path accepted an existing config that way.

## Why derive rather than validate

Issue #105 suggested enforcing `is_new => data.is_empty()` at decode. This PR goes further and drops the flag, because the flag carries no information a verifier could check against:

- Writing empty data produces an `EMPTY_HASH` commitment, and the SMT updater removes the leaf (`core/smt/src/updater.rs:129`).
- An absence proof reports its value hash as `EMPTY_HASH` too (`core/smt/src/proving/member.rs:16`).

So "absent" and "present but empty" are the same state in the SMT, by construction. Committing `is_new` in the journal would leave the verifier with only one available check, against `value_hash == EMPTY_HASH`, which is exactly `data.is_empty()`. `InputResourceCommitment::encode` already hashes `[] => EMPTY_HASH` and the verifier already compares that hash against the proof, so emptiness is bound today. A wire flag is redundant state that can only diverge from what is bound.

Newness is derived at decode as a snapshot, not as a live `data().is_empty()` accessor, so a guest that creates a slot and writes to it still reads `is_new() == true`.

## Lifecycle consequence

Committed state has no deleted-but-present slot, so `Lifecycle::from_resource` starts every empty slot at `New` and `Deleted` is only reached by a transaction's own delete transition (which still blocks the within-tx double-deposit path, #92).

The removed committed-empty to `Deleted` arm previously rejected re-creation, but only against an honest host: a prover flipped the bit and got `New` regardless. Note the host already reported a previously-emptied resource as `is_new=true` across batches (`write_latest_ptr` deletes the pointer when data is empty), so this makes the cross-batch and within-batch behavior consistent rather than changing an enforceable rule.

`AccessHandle::is_new` (the host-side `version() == 0` source of the bit) is removed as unused.

## Tests

- `init_over_live_config_rejected` (the #105 repro) passes.
- Adds `init_creates_config_when_slot_is_empty`: the identical genesis-authorized `Init` over an empty slot succeeds, pinning the rejection to the slot's lifecycle rather than the auth path or action encoding.
- Adds ABI unit tests for empty-decodes-as-new, non-empty-decodes-as-live, and the snapshot-survives-writes property.
- Guest ELFs rebuilt for the new wire format; without that, the runtime-processor e2e suite fails wholesale on a wire mismatch.

Verified: 28/28 runtime-processor e2e, 95/95 runtime-processor lib, 264/264 workspace (nextest), workspace and runtime-processor clippy clean under `-D warnings`.


zk/backend/risc0/runtime-processor/tests/e2e.rs
```diff
@@ -770,6 +771,81 @@ fn prev_tx_v1_witness_unlocks_schnorr_locked_config() {
...
+
+    let tx_bytes = encode_v1_transaction(&payload, &current_rest_preimage);
+    encode_inputs(0, [0u8; 32], &tx_bytes, &[(0, config_data)])
+}
+
+/// A live, non-empty config cannot be re-initialized. Newness is derived from the resource data,
+/// which the journal's data hash binds to the old root, so a prover cannot present these committed
+/// bytes as a new slot; `Init` sees `Live` and must reject.
+#[test]
+fn live_config_falsely_marked_new_is_rejected() {
```
hmoog: `live_config_falsely_marked_new_is_rejected` is named after the old attack, but nothing "marks" newness anymore. Optional rename: `init_over_live_config_rejected`.

scheduling/scheduler/src/access_handle.rs
hmoog: `version()` doc still says "(0 if new)". Harmless, but "new" is no longer a host-side concept after removing `is_new()`.

hmoog (approved):
Just two minor optional changes. Otherwise its a good change that aligns the behavior of the underlying storage layer with the treatment in the guest.

PR #118, opened 2026-07-17
Canonicalize prover-supplied structure in the batch verifier

### LINKED ISSUE #106: Batch verifier does not canonicalize prover-supplied proof structure
## Summary

`batch_processor::Verifier` does not canonicalize the prover-supplied SMT proof or the tx journal structure. The proof and journals are private prover data, so their shape is untrusted, and the verifier is the only place that can constrain it.

A prover can append a duplicate key/membership pair to an honest proof so that two withdrawals alias one account key. The aliased entry is invisible to the covenant, which sees only the state roots.

## Impact

Two payouts totalling 2000000 settle against a balance funding a single payout of 1000000, while only one debit reaches the state root. The batch settles instead of being rejected.

## Reproduction

Repro branch:

https://github.com/biryukovmaxim/vprogs/tree/repro/g1-batch-verifier-proof-canonicalization

Commit:

`e9b28c3 Add batch-verifier canonicalization repros`

Focused command:

```bash
cargo test -p vprogs-zk-abi --test batch_processor
```

```text
batch settled instead of being rejected
  exits: 2 payouts totalling 2000000 against a balance funding one payout of 1000000
  prev_state: e5c75ad1... (on-chain root e5c75ad1..., honest proof e5c75ad1...) -- alias invisible to covenant
  new_state:  9bbeb9d2... (root folding a single debit) -- one debit reached the root, two were paid
```

## Root Cause

The malicious proof is the honest proof's own wire bytes with `keys[0]` and `memberships[0]` appended. The verifier accepts the duplicate rather than rejecting a non-canonical proof, so two withdrawals resolve through one membership index.

The tests use a real `RocksDbStore` SMT and a real `Tree::prove` proof. Each `verifier_rejects_*` test states a property the verifier must hold and fails against a verifier that does not hold it.

The control test is load-bearing: with both transactions at `resource_index = 0` the verifier **does** panic with `resource hash mismatch`, which proves the chaining works and that the alias is what defeats it.

## Scope

- No RISC0 run. `verify_batch_journal` is a no-op closure: these tests exercise the verifier's own shape checks, not `env::verify`.
- The payout is inferred from `BatchTransition.exits` rather than observed on chain.
- The `Inputs` wire bytes are hand-encoded.
- The only non-test change is three dev-dependencies in `zk/abi/Cargo.toml`.

## Suggested Fix

Canonicalize prover-supplied structure in the verifier: reject a proof whose keys are not strictly ordered and deduplicated, and reject a success journal with more outputs than input resources. The repro's three `verifier_rejects_*` tests passing is the acceptance criterion.

## Related

**Related to #105** (runtime trusts unbound `is_new` resource flag). The two share `verified_batch_idx` and the `batch_processor::Verifier` surface, and #105's own hardening notes observe that "resource index is journaled and checked, but depends on proof canonicality" — this issue is that dependency. The `Verifier` harness #105 lacked now exists on this branch. Neither blocks the other.

biryukovmaxim:
## Summary

Fixes #106.

The SMT proof and tx journals are prover-supplied private data, so their shape is untrusted and `batch_processor::Verifier` is the only place that can constrain it. This PR adds the missing shape checks, with the repro tests from the issue as the acceptance criterion.

## Changes

- **Canonical batch resource order.** The scheduler now assigns batch-local resource indices in ascending resource-id order (pre-scanning the batch's declared accesses into an id-to-index map) instead of first-touch order, and state diffs are sorted so positions match indices. Access lists were already strictly ascending per transaction, enforced at payload decode; this extends the same canonical order to the batch, so the proof's queried keys arrive sorted.

- **Reject a proof whose queried keys are not strictly ascending.** A membership at index `i` always queries `keys[i]`, so a duplicated key gives one resource two independent membership slots: two withdrawals alias one account past the per-resource hash chaining while only one debit reaches the state root. With the canonical order above, `Proof::check_sorted_queried_keys` enforces sortedness and uniqueness in one alloc-free `is_sorted_by` pass, mirroring the per-tx `AccessMetadata::decode_slice` rule.

- **Reject a success journal with more output commitments than declared input resources.** Trailing output commitments were never read, so a journal could commit output state the verifier ignored.

- **Rebuilt guest ELFs.** The repository ships pre-built guest binaries, so the new checks only reach the executable circuit through a rebuild (`./zk/backend/risc0/build-guests.sh`). The batch-processor, both transaction-processor variants, and the runtime-processor ELFs changed; batch-aggregator rebuilt byte-identical, confirming the Docker build is deterministic. The rotated image ids also move the proof-receipt cache (`TxKey`/`BatchKey` embed the image id) to a fresh namespace, so pre-upgrade receipts journaling first-touch indices become cache misses instead of replaying against sorted proofs.

- **Result-based test harness.** The `batch_processor` tests' `settle` helper catches the verifier's rejection unwind and returns `Result<Settled, String>`, so tests assert on `Err` values (printing the actual message on mismatch) instead of `should_panic` attributes and inline `catch_unwind` blocks.

## Testing

- The `batch_processor` repro tests pass: the aliasing and extra-outputs attacks are rejected, the control test still fails with `resource hash mismatch` (showing the hash chaining is what the alias was defeating), and a new test rejects a proof querying keys out of canonical order.
- A new scheduler e2e test drives first-touch order 5, 1, 9 and asserts indices come back assigned 1, 5, 9 with state-diff positions matching.
- 98 tests across `vprogs-core-smt`, `vprogs-zk-abi`, and the scheduler pass; the dev-mode RISC0 suite (24 e2e tests) and the sim crate (8 seeded L2 flows, including reorgs) were re-run against the rebuilt guests, so the checks demonstrably execute inside the real circuit.
- A Codex review of the branch raised two P1s, both addressed by the ELF rebuild: the stale committed batch-processor guest, and pre-upgrade receipt replay (resolved by the image-id rotation in the receipt keys).


zk/abi/tests/batch_processor.rs
```diff
@@ -0,0 +1,406 @@
...
+        &Hash::from_bytes([0xE2; 32]),
+        1,
+        &context_hash,
+        &[(2, ALICE_KEY, ALICE_FUNDED)],
+        &withdraw,
+        &[Some(ALICE_DEBITED)],
+    );
+    let input_bytes = batch_inputs(&evil_proof, &[tx1, tx2]);
+
+    let Ok(settled) = settle(&input_bytes) else {
```
hmoog: The aliasing repro accepts any rejection (`let Ok(settled) = settle(..) else { return }`). If the hand-rolled `tx_journal`/`batch_inputs` encoders ever drift from the real wire format, `settle` fails on decode and the test silently passes while exercising nothing. Pin the expected message ("canonical proof keys") the way `verifier_rejects_unsorted_queried_keys` pins its rejection.

zk/abi/tests/batch_processor.rs
```diff
@@ -0,0 +1,406 @@
...
+        0,
+        &context_hash,
+        &[(0, ALICE_KEY, ALICE_FUNDED)],
+        &[],
+        // One declared input resource, three output commitments.
+        &[Some(ALICE_DEBITED), Some([0xCC; 32]), None],
+    );
+    let input_bytes = batch_inputs(&proof, &[tx]);
+
+    if settle(&input_bytes).is_err() {
```
hmoog: Same for the extra-outputs repro: assert the rejection contains "more output commitments than input resources" instead of accepting any error.

hmoog (approved):
Looks good as well, just two comments in the tests, to enforce the exact invariant we are testing.

Issue #111, opened 2026-07-16
Receipt-cache keys omit the pins that define the proved statement


biryukovmaxim:
## Summary

The proof-receipt cache keys do not name the values that define the statement they key.

`BatchKey` (checkpoint index, block hash, batch image id) omits all four `BatchPins` (`tx_image_id`, `covenant_id`, `deposit_spk_hash`, `lane_key`) that `zk/abi/src/batch_processor/verifier.rs:99-114` commits into the batch transition. `AggregatorKey` additionally omits `lane_key` and the batch image id.

## Impact

This key-domain defect is **unconditional**: on a cache hit, `zk/batch-prover/src/worker.rs:90-91` and `zk/aggregate-prover/src/worker.rs:294-295` publish the stored receipt without checking any committed field against current configuration, so a prover is handed a cryptographically valid proof of a statement it did not ask for.

Separately and also unconditionally, the only binding of the aggregate receipt's committed `new_seq_commit` to the scheduled block, the value the key itself claims, is a `debug_assert_eq!` at `zk/aggregate-prover/src/worker.rs:350`. It is compiled out in release and placed 26 lines **after** the `write_agg_receipt` at `:324`. The adjacent `covenant_id` check at `:343-348` is a real `assert_eq!`, so the weakening is deliberate.

The resulting **denial of service is conditional**: reuse-across-pins requires two lanes or configurations to share a receipt store **and** the same checkpoint/block/image coordinates, normally a config or lane change surviving a restart. The poisoning wedge requires a receipt whose committed seq_commit diverges.

The on-chain covenant's embedded pins prevent either from becoming an accepted forged transition, confining impact to DoS.

## Reproduction

Repro branch:

https://github.com/biryukovmaxim/vprogs/tree/repro/g6-receipt-cache-pins

Focused commands:

```bash
cargo test -p vprogs-zk-aggregate-prover --test receipt_cache_pin_omission -- --test-threads=1

RUSTFLAGS="-C debug-assertions=off" cargo test -p vprogs-zk-aggregate-prover \
    --test receipt_cache_poisoning -- --test-threads=1
```

Observed result on current `master`, 10/10 runs each:

```text
assertion `left == right` failed: the artifact handed to lane B's settlement must prove lane B's
transition, not be lane A's receipt served from a key that omits the lane
  left: a0a0...a0   right: b0b0...b0

a receipt committing a new_seq_commit other than the scheduled final block's must be rejected before
[...]

PR #120, opened 2026-07-17
Classify no-op bundles on the full committed transition

### LINKED ISSUE #109: Aggregate wrapper discards bundles that are provably not no-ops
## Summary

The aggregate prover's wrapper classifies a bundle as a no-op on SMT state-root equality alone (`zk/aggregate-prover/src/worker.rs:338-341`), on the premise, stated in the adjacent comment at `:335-337`, that an unchanged root means the bundle's blocks carried no lane activity.

That premise is false. The batch guest folds every transaction into the lane activity digest including failed ones (`zk/abi/src/batch_processor/verifier.rs:165-170`), and any batch with at least one tx journal derives a fresh lane tip from that digest (`:50-83`), while an `OutputCommitment::Error(_)` tx performs no resource write (`:134-140`) and `latest_value_hashes` updates only on `Changed` (`:156-161`, `:93-97`), so the root is carried through unchanged.

State root and lane tip move independently, and `StateTransition` commits them independently (`zk/abi/src/batch_aggregator/journal/state_transition.rs:8-36`).

## Impact

The only earlier guard filters on tx count, not state change (`:258`), so a bundle containing a single ordinary failed transaction reaches the branch, is consumed, and publishes `None` while its lane tip, exits, and deposit hash are discarded (`:359-371`).

The covenant is then never advanced, the next bundle correctly chains `prev_lane_tip` to the advanced tip, and the settler's equality at `zk/backend/risc0/settler/src/covenant/build.rs:128-131` rejects it and every bundle after it. Settlement liveness is lost permanently, with no adversary required.

## Reproduction

Repro branch:

https://github.com/biryukovmaxim/vprogs/tree/aggregate-noop-full-transition

Focused commands:

```bash
cargo test -p vprogs-zk-aggregate-prover --test noop_bundle_classification
cargo test -p vprogs-zk-backend-risc0-test-suite --test settlement_chain_wedge
```

```text
thread 'bundle_advancing_only_the_lane_tip_is_settled' panicked at zk/aggregate-prover/tests/noop_bundle_classification.rs:385:42:
a bundle whose lane tip advanced must be settled, not discarded as a no-op: its covenant is never
advanced and every later bundle's prev_lane_tip then chains from a tip the covenant does not carry
```

Three separate paths publish no artifact (the tx-count guard, the shutdown branch at `worker.rs:318`, and the target state-root branch at `worker.rs:339`). The test asserts the bundle was actually proved before asserting the artifact is `None`, which pins `:339` as the cause rather than an earlier return.

`aggregator_commits_unchanged_state_with_an_advanced_lane_tip` runs the real `Verifier` and confirms `new_state == prev_state` while `new_lane_tip != prev_lane_tip`. It passes and must keep passing.

The settler wedge at `covenant/build.rs:128-131` is genuinely executed: `bundle_after_a_dropped_one_cannot_settle` calls the real `build_settlement`.

## Scope

- Real L1 consensus was **not** exercised. No node, no chain, no block submission.
- No ZK proving occurred. The aggregator backend is stubbed, the settler uses a `FakeReceipt`, and image ids are placeholders.
- The batch journal is hand-encoded via `BatchTransition::encode` rather than produced by a real batch guest run. The aggregator verifier is real; the batch guest is not executed.
- The settler tests prove the wedge against the settler's own precondition, not against a live chain, and cover only the first bundle after the drop. Permanence is argued, not enumerated.

## Related

**Conflicts with #111** (receipt-cache keys omit the pins that define the proved statement). Both edit `zk/aggregate-prover/src/worker.rs` on different predicates, so the two fixes conflict textually and should be sequenced. Neither blocks the other.

## Suggested Fix

Classify a bundle as a no-op on the full committed transition (state root **and** lane tip, exits, deposit hash), not on state-root equality alone. Un-ignoring is not required; the repro test failing is the acceptance criterion and it passes once the wrapper stops discarding tip-advancing bundles.

biryukovmaxim:
## Summary

The aggregate wrapper resolved a bundle as a no-op whenever its state root was unchanged (`zk/aggregate-prover/src/worker.rs`). That premise is false: the batch guest folds every transaction into the lane activity digest, including failed ones, so a batch whose only transaction failed carries the state root through unchanged while still advancing the lane tip. Such a bundle was consumed and published as `None`, its lane tip, exits, and deposit hash discarded. The covenant then never advanced, the next bundle correctly chained `prev_lane_tip` past the covenant's stale tip, and the settler's chain check rejected it and every bundle after it: settlement liveness lost permanently, no adversary required.

The fix classifies a no-op on the whole committed transition: state root, lane tip, exits (`permission_spk_hash`), and deposit hash must all be unchanged before the wrapper resolves the handle without settling.

## Commits

The first three commits are the repro (previously on `repro/g4-aggregate-wrapper-drops-nonnoop-bundles`, now folded into this branch):

- `noop_bundle_classification` runs the real aggregator `Verifier` and pins that `new_state == prev_state` can coexist with `new_lane_tip != prev_lane_tip`, then drives the real worker and asserts the bundle is settled rather than dropped. It fails without the fix and passes with it.
- `settlement_chain_wedge` calls the real `build_settlement` and pins the downstream consequence: once one advance is dropped, the next bundle cannot settle.

## Testing

- `cargo nextest run -p vprogs-zk-aggregate-prover -p vprogs-zk-backend-risc0-test-suite`: 35/35 pass (the repro test fails on master without the fix commit).
- Downstream consumers (`vprogs-sim` incl. the reorg/settlement l2-flow tests, `vprogs-zk-vm`, settler, api): 23/23 pass.
- fmt, clippy, and `./cargo-check-all.sh` clean.

Note: conflicts textually with the fix for #111 (both touch `worker.rs` on different predicates); whichever lands second needs a trivial rebase.

Fixes #109


.env
```diff
@@ -0,0 +1,10 @@
+# Local secrets for the tn10 demos. Gitignored — do NOT commit.
```
hmoog: This file must not be in the PR. It carries two funded testnet private keys and a node endpoint, its own header says "do NOT commit", and it slipped in via the merge commit (af2a97c), whose message itemizes every other change but not this one. Since the branch is already pushed to a public fork, treat the keys as compromised: rotate them and move any tn10 funds, then rewrite the branch so the file is not in its history (deleting it in a follow-up commit still publishes it via history), and add `.env` to `.gitignore` so this cannot recur. Nothing in this PR reads these variables, so removal costs nothing.

hmoog (approved):
Only one comment regarding the .env file

PR #121, opened 2026-07-17
Bridge worker: surface malformed RPC responses as Fatal instead of panicking

### LINKED ISSUE #115: Bridge worker panics on RPC-wire data with no Fatal event
## Summary

`BridgeWorker::fetch_chain_updates` and `lane_state` (`l1/bridge/src/worker.rs:459,460,470,505,526,543`) `expect`/`unwrap` fields that the Kaspa wRPC peer carries as `Option`: `hash`, `daa_score`, `blue_score`, `timestamp`, and the transaction's `storage_mass`.

The worker runs on a bare `thread::spawn` (`l1/bridge/src/bridge.rs:51-58`) with no `catch_unwind` anywhere in the crate (`grep -rn catch_unwind l1/bridge/src/` returns nothing).

## Impact

A peer that omits any one of them panics the worker thread. The panic never becomes an `L1Event::Fatal`, so the sink stops advancing, no event is queued, and consumers parked in `wait_and_pop` block forever with nothing observable to distinguish it from a quiet chain.

Since `L1BridgeConfig::default()` leaves `url: None` (`config.rs:53`) and `worker.rs:105` turns that into the public community resolver, the peer is untrusted and one elided field is a remote liveness kill.

## Reproduction

Repro branch:

https://github.com/biryukovmaxim/vprogs/tree/repro/g10-bridge-worker-rpc-panics

Focused command:

```bash
cargo test -p vprogs-l1-bridge --test wire_panic_repro -- --test-threads=1
```

A tampering wRPC proxy in front of a real simnet node elides one field per case. Observed result on current `master`:

```text
thread '<unnamed>' panicked at l1/bridge/src/worker.rs:459:42: missing hash
thread 'elided_header_hash_reports_fatal' panicked at l1/bridge/tests/wire_panic_repro.rs:372:5:
worker.rs:459: expected Fatal on the elided field, got Err(Elapsed(())). The worker scheduled 0
blocks, and panicked at: ["panicked at l1/bridge/src/worker.rs:459:42:\nmissing hash"]

thread '<unnamed>' panicked at l1/bridge/src/worker.rs:460:46:  missing daa_score
thread '<unnamed>' panicked at l1/bridge/src/worker.rs:470:66:  missing tx fields: MissingRpcFieldError("RpcTransaction", "storage_mass")
thread '<unnamed>' panicked at l1/bridge/src/worker.rs:505:60:  called `Result::unwrap()` on an `Err` value: "missing timestamp"
thread '<unnamed>' panicked at l1/bridge/src/worker.rs:526:44:  missing blue_score
```

The full chain is executed: real simnet kaspad -> tampering proxy eliding one field -> real `L1Bridge` on its real `thread::spawn` worker -> panic at the named line -> `wait_and_pop` returns `Err(Elapsed)` after 15s -> sink frozen at 0 blocks and no event queued after 5 further blocks are mined.

The suite needs a real kaspad per case and takes ~97s at `--test-threads=1`.

## Five sites, not six

`:543` is **unreachable, not missed**, and has no case. `lane_state` is called at `:490` with `header = &chain_block.chain_block_header`, the same reference `:460` already `expect`ed `daa_score` on, earlier in the same loop iteration. No response can reach `:543` while `:460` stands. It is doubly unreachable under the default config: `lane_key` is `None` (`subnetwork_id: None`), so `:530` returns early.

**A `:460`-only fix leaves `:543` live.** It must be fixed alongside and caught by review, since no test can cover it.

## Correction to the finding's supporting argument

`get_vcc_with_retry` (`worker.rs:398-428`) exists, but it retries on `Err` from the RPC **call** (transport and `is_fatal` classification) and has no notion of a missing `Option` field. The two sites guard different things, so the apparent "inconsistency within the same file" is not evidence for this defect. The defect is real regardless; that particular argument for it is not.

## Scope

- **Traced, not executed:** that the public community resolver actually reaches an untrusted peer in production. Verified at source (`config.rs:53`, `worker.rs:105`); no real public peer was driven.
- The assert demands `Fatal`, which is stronger than "don't panic". `L1Event` has only `Connected`/`Disconnected`/`Fatal`, and `handle_sync_result` log-and-retries non-fatal errors, so a fix classifying a malformed response as non-fatal would retry a peer that can never serve a usable response, which is still a silent hang. These tests would still fail in that case. That matches the stated contract but is a design choice the fix inherits.
- `malformed_response_leaves_the_bridge_observable` asserts progress **or** an event, so it stays valid after any reasonable fix.

## Suggested Fix

Treat the wire `Option`s as untrusted input: return a classified error rather than `expect`/`unwrap`, and surface it as `L1Event::Fatal` so parked consumers observe it. Fix `:460` and `:543` together.

## Related

**Touches the same file as #113** (finalization never advances on a live node): both change `l1/bridge/src/worker.rs`, but different functions with no shared test surface, so they should not conflict. Neither blocks the other.

biryukovmaxim:
Closes #115.

## Problem

`BridgeWorker::fetch_chain_updates` and `lane_state` `expect`/`unwrap`ed fields the wRPC peer carries as `Option` (`hash`, `daa_score`, `blue_score`, `timestamp`, and the transaction's `storage_mass`). The worker runs on a bare `thread::spawn`, so a peer eliding one field panicked the worker thread with no `L1Event::Fatal`: the sink froze and consumers parked in `wait_and_pop` blocked forever, indistinguishable from a quiet chain. `L1BridgeConfig::default()` reaches the public community resolver, so the peer is untrusted and one elided field was a remote liveness kill.

## Fix

- New fatal `Error::MalformedResponse` variant. It is mapped explicitly rather than through the existing `From<RpcError>` conversion, which would classify it as recoverable `Error::Rpc` and log-and-retry a peer that can never serve a usable response (still a silent hang).
- `fetch_chain_updates` validates each chain-block header once up front via the existing `ChainBlockMetadata::try_from` (which already checks all five required header fields) and propagates `MalformedResponse` for elided header or transaction fields. The error flows through `handle_sync_result` into `L1Event::Fatal`, waking parked consumers.
- `lane_state` now takes the validated `ChainBlockMetadata` instead of the raw wire header, so its own `Option` reads (including the unreachable `daa_score` one the issue flagged for review) are gone structurally.
- `seed_from_recent`'s selected-parent walk returns `MalformedResponse` instead of `expect`ing `verbose_data`. That one runs during chain init, before `L1Event::Connected` is pushed, so the panic left a consumer with no event at all; `handle_connected` already routes chain-init errors into `Fatal`. The other two `get_block` calls read only the non-optional `.header`.

## Tests

The wire-panic repros (first two commits) become malformed-response contract tests, renamed to `malformed_response.rs`: a tampering wRPC proxy in front of a real simnet node elides one field per case, and each case asserts `wait_and_pop` returns `L1Event::Fatal`. The proxy picks the reply it rewrites from the field under test, so the seed case elides `verbose_data` from a `GetBlock` reply and spawns its harness without waiting for `Connected`. Each case fails with `Err(Elapsed)` before the fix; all 7 pass after (in ~12s instead of ~97s, since `Fatal` fires immediately). Bridge suite: 13/13. Workspace check and clippy clean.

Follow-up (#123, from review): `MalformedResponse` is terminal, so a byzantine peer handed out by the public resolver can end the bridge for good where rotating to another peer would preserve liveness. Validation happens before any append, so the sink prefix stays consistent and a resync from tip would continue cleanly. Left as-is here; the hardening run owns it.

### DIFF
--- a/l1/bridge/src/error.rs
+++ b/l1/bridge/src/error.rs
@@ -16,6 +16,10 @@ pub(crate) enum Error {
     #[error("reorg below finalization boundary: fork block {0} is finalized")]
     ReorgBelowFinality(Hash),
 
+    /// The peer elided a response field required at `Full` verbosity - retrying cannot help.
+    #[error("malformed RPC response: {0}")]
+    MalformedResponse(String),
+
     /// An internal channel closed unexpectedly.
     #[error("notification channel closed: {0}")]
     ChannelClosed(String),
--- a/l1/bridge/src/worker.rs
+++ b/l1/bridge/src/worker.rs
@@ -13,7 +13,6 @@ use kaspa_notify::scope::{PruningPointUtxoSetOverrideScope, Scope, VirtualChainC
 use kaspa_rpc_core::{
     GetVirtualChainFromBlockV2Response, Notification,
     RpcDataVerbosityLevel::Full,
-    RpcOptionalHeader,
     api::{ctl::RpcState, rpc::RpcApi},
 };
 use kaspa_seq_commit::{
@@ -453,41 +452,42 @@ impl<T: ChainSink<ChainBlockMetadata, L1Transaction>> BridgeWorker<T> {
 
         // Schedule each new block, threading its parent from the current tip.
         for chain_block in response.chain_block_accepted_transactions.iter() {
+            // Validate the peer-controlled header up front; an elided field fails the sync and
+            // downstream code sees only the validated metadata.
+            let block = ChainBlockMetadata::try_from(&chain_block.chain_block_header)
+                .map_err(|field| Error::MalformedResponse(field.into()))?;
+
             // The block's parent; its `last_settlement` carries forward, updated per tx below.
-            let header = &chain_block.chain_block_header;
             let parent = self.tip_metadata();
-            let block_hash = header.hash.expect("missing hash");
-            let block_daa = header.daa_score.expect("missing daa_score");
             let mut last_settlement = parent.last_settlement;
 
             // Enumerate before filtering so kept txs keep their block-wide positions.
-            let txs: Vec<SchedulerTransaction<L1Transaction>> = chain_block
-                .accepted_transactions
-                .iter()
-                .enumerate()
-                .filter_map(|(idx, tx)| {
-                    // Carry forward the last settlement.
-                    let tx = L1Transaction::try_from(tx.clone()).expect("missing tx fields");
-                    if let Some(id) = self.covenant_id {
-                        last_settlement =
-                            tx.settlement_info(id, block_hash, block_daa).or(last_settlement);
-                    }
+            let mut txs: Vec<SchedulerTransaction<L1Transaction>> =
+                Vec::with_capacity(chain_block.accepted_transactions.len());
+            for (idx, tx) in chain_block.accepted_transactions.iter().enumerate() {
+                // Same trust boundary as the header: elided tx fields fail the sync.
+                let tx = L1Transaction::try_from(tx.clone())
+                    .map_err(|e| Error::MalformedResponse(e.to_string()))?;
 
-                    // Parse access metadata; marlformed = no dependencies and prover attests.
-                    match self.subnetwork_filter.as_ref() {
-                        Some(want) if tx.subnetwork_id != *want => None,
-                        _ => Some(SchedulerTransaction::new(
-                            idx as u32,
-                            AccessMetadata::decode_vec(&mut tx.payload.as_slice())
-                                .unwrap_or_default(),
-                            tx,
-                        )),
-                    }
-                })
-                .collect();
+                // Carry forward the last settlement.
+                if let Some(id) = self.covenant_id {
+                    last_settlement =
+                        tx.settlement_info(id, block.hash, block.daa_score).or(last_settlement);
+                }
+
+                // Parse access metadata; malformed = no dependencies and prover attests.
+                match self.subnetwork_filter.as_ref() {
+                    Some(want) if tx.subnetwork_id != *want => {}
+                    _ => txs.push(SchedulerTransaction::new(
+                        idx as u32,
+                        AccessMetadata::decode_vec(&mut tx.payload.as_slice()).unwrap_or_default(),
+                        tx,
+                    )),
+                }
+            }
 
             // Determine the lane tip over the block's accepted txs.
-            let (lane_tip, lane_blue_score, lane_expired) = self.lane_state(&parent, &txs, header);
+            let (lane_tip, lane_blue_score, lane_expired) = self.lane_state(&parent, &txs, &block);
 
             // Append the block's parent-threaded metadata and its txs to the sink.
             self.sink.append(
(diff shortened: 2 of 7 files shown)


l1/bridge/src/worker.rs
hmoog: `seed_from_recent` still panics on the same trust boundary: `.verbose_data.expect("get_block returns verbose data")` consumes an optional field of a `get_block` reply from the same untrusted connection. A peer eliding it crashes the worker at startup exactly the way the fixed sites did during sync. Convert it to `Error::MalformedResponse` like the rest.

l1/bridge/src/error.rs
```diff
@@ -16,6 +16,10 @@ pub(crate) enum Error {
     #[error("reorg below finalization boundary: fork block {0} is finalized")]
     ReorgBelowFinality(Hash),
 
+    /// The peer elided a response field required at `Full` verbosity - retrying cannot help.
```
hmoog, line 19: "Retrying cannot help" is true for the same peer, but with the default config the peer comes from the public resolver, so a single byzantine node can now terminate the bridge for good, where disconnecting and resolving a different peer would preserve liveness (the sink prefix is consistent, since validation happens before any append of the failing block, so a resync from tip would continue cleanly).

Fatal-with-event is strictly better than the panic and fine for this PR - but maybe we should create an issue already where we plan to revisit the retry strategies later in a hardening run.

hmoog (changes requested):
Two small things,

l1/bridge/src/worker.rs
biryukovmaxim: Fixed in 947fc72. `seed_from_recent`'s walk now returns `Error::MalformedResponse("missing block verbose_data")`; `handle_connected` already routes chain-init errors into `L1Event::Fatal`, so the seed path surfaces the same way the sync path does.

Worth noting it was the worse of the two: the walk runs before `L1Event::Connected` is pushed, so the panic left a consumer with no event at all, not even the connect.

Covered by a new case, `elided_block_verbose_data_reports_fatal`. The tampering proxy now picks the reply it rewrites from the field under test, so it can elide `verbose_data` from a `GetBlock` reply, and the harness spawns without waiting for `Connected`. Against the old code it fails with `Err(Elapsed)` and records the panic at `worker.rs:333`; 7/7 pass now.

The two remaining `get_block` calls (`seed_from_pruning_point`, `seed_from_block`) only read `.header`, which is not optional.

l1/bridge/src/error.rs
biryukovmaxim: Agreed on both points, and the clean-prefix argument holds: validation runs before any append, so the sink is consistent at the moment of the error and a resync from tip continues with no rollback and no gap.

Filed #123 for the hardening run: reclassify `MalformedResponse` as "not retryable against this peer, but recoverable by replacing it", rotate the resolver-supplied peer instead of terminating, bounded so a chain of bad peers cannot spin forever, and keep `Fatal` when there is nothing to rotate to (explicit `url`) or once the bound is exhausted. Covers the seed path too, which the latest commit brings under the same error.

Leaving the behavior in this PR as it is.

hmoog (approved):
Approved

Issue #103, opened 2026-07-15
l1/wallet: make funded tx builders fallible instead of panicking on insufficient funding


biryukovmaxim:
The `build::*` transaction builders in `l1/wallet` panic via `assert!` when the chosen funding UTXO can't cover the outputs plus the node's minimum mass-based fee:

- `build::signed_carrier_transaction` — `amount > extra_value`, `amount > extra_value + fee`
- `build::pay_to_address_transaction` — `amount > payout`, `amount > payout + fee`
- `build::activity_transaction` — `amount > fee`
- `build::settlement_transaction` — `fee_entry.amount > fee`

These are `pub` and used in production through `Wallet`, which selects the UTXO to fund from (`build_signed_carrier`/`build_covenant_bootstrap_transaction` take `utxos.into_iter().next()`; `pay_to_address` the largest). A UTXO smaller than the required outputs+fee is a reachable *runtime* condition, not a programmer error, so it should surface as a recoverable error rather than crash the issuer.

### Proposed change

- Add a `thiserror` `BuildError` (e.g. `InsufficientFunds { available, required }`) in `l1/wallet/src/build.rs`; have the builders above return `Result<Transaction, BuildError>`. Fold the zero-fee probe's underflow guard into a single post-probe sufficiency check by probing the change output with a placeholder value (its value doesn't affect the byte layout / mass), the way `activity_transaction` already does.
- Propagate through the `Wallet` methods (`build_payload_transactions`, `build_subnet_payload_transactions`, `build_signed_carrier`, `build_covenant_bootstrap_transaction`, `prepare_settlement_transaction`/`_excluding`, `pay_to_address`, `build_activity_excluding`) via a `WalletError` folding `RpcError` + `BuildError` + a no-spendable-UTXO case; the fetch `.expect(...)`s become `?`.
- Update callers (settler, `examples/tn10-flow`, `sim` driver, `node/test-utils`, e2e tests).

`build::covenant_bootstrap_transaction` does no fee math and has no funding-insufficiency path, so its builder stays infallible; only its `Wallet` wrapper gains the UTXO-selection error.

biryukovmaxim:
Partially addressed by #122: \`activity_transaction\`, \`pay_to_address_transaction\` and \`settlement_transaction\` now fund from a candidate-UTXO list and return a typed \`BuildError\` (\`InsufficientFunds\`, \`StorageInfeasible\`) instead of asserting, propagated through the \`Wallet\` methods and their callers.

[...]

PR #122, opened 2026-07-20
Price wallet fees on the node's floor and fee estimate; fund from candidate-UTXO prefixes

### LINKED ISSUE #116: Wallet fees are priced on the wrong masses: storage charged, transient omitted
## Summary

`min_fee` (`l1/wallet/src/build.rs:33`) prices every builder at
`MIN_FEERATE_PER_GRAM * max(compute_mass, storage_mass)`. Both halves of that formula are wrong
against the node the wallet submits to, and every builder in `l1/wallet/src/build.rs` inherits
both errors (`activity_transaction:100-110`, `pay_to_address_transaction:169-180`,
`settlement_transaction:273-277`).

The node's admission rule (`mining/src/mempool/check_transaction_standard.rs:127-158` at the
pinned rusty-kaspa rev `9fdbaf1`) is:

```rust
// Minimum relay fee applies to normalized non-contextual mass so block-space usage has a
// minimum cost, whether dominated by compute or by transient byte footprint.
// Storage mass does not require an additional relay-fee floor here since storage growth is
// sufficiently protected even under worst-case block-limit usage.
let (fee_mass, relay_fee) = ... (masses.compute_mass.max(normalized_transient_mass),
                                 self.config.minimum_relay_transaction_fee);
```

**Storage mass is charged but the node never requires it.** This is the source of the
non-convergence: paying the fee shrinks the change output, KIP-0009's `C/v` term makes a smaller
output cost *more* storage mass, so `min_fee(final)` strictly exceeds the fee the `fee = 0` probe
was priced for, and each builder's `assert!` cannot catch it because it prices the probe, not the
result. The fixpoint only exists because a mass the node does not price was put in the fee
formula.

**Normalized transient mass is required but never charged.** Transient mass is 2 grams per
serialized byte against compute's ~1, so byte-heavy shapes -- a payload-carrying activity
transaction, a witness-heavy settlement -- pay below the node's floor and are rejected with
`RejectInsufficientTransientFee`.

## Reachability

An earlier revision of this issue argued the underpayment was latent, on the grounds that the
mempool priced compute alone at `DEFAULT_MINIMUM_RELAY_TRANSACTION_FEE = 1000` sompi/kg = 1
sompi/gram against the wallet's `MIN_FEERATE_PER_GRAM = 100`, a ~100x cushion. **That no longer
holds at the pinned rev.** `DEFAULT_MINIMUM_RELAY_TRANSACTION_FEE` is now `100_000` sompi/kg = 100
sompi/gram, exactly `MIN_FEERATE_PER_GRAM` (`mining/src/mempool/config.rs:23`); the old 1000 value
survives only as `LEGACY_MINIMUM_RELAY_TRANSACTION_FEE`, applied to low-priority submissions
before Toccata activation. There is no cushion left: any shortfall against the node's floor is a
rejection, so the transient-mass underpayment is directly reachable today.

## Storage mass is a funding constraint, not a fee term

Removing storage from the fee does not make it irrelevant. It gates validity in two ways no fee
can buy past:

- the block-fit limit (`block_mass_limits().raw_post().storage`), and
- per-output viability: below roughly 0.02 KAS the `C/v` term makes a change output impossible to
  create at all.

The builders have no answer to the second. Each does one probe and one rebuild, with no way to
drop an unviable change output or pull in another input, and asserts rather than returning an
error when the arithmetic does not work out.

## The fixpoint that is real

Landing a transaction competitively *is* a fixpoint, but at a different layer than the floor. The
mempool orders by fee over `Mass::normalized_max` of compute, transient, **and** storage
(`FeerateTransactionKey::from_tx`), so a fee targeting an estimator's rate does depend on the
change value it is deducted from: raising the fee shrinks the change, which raises storage mass,
which raises the required fee. The floor, once priced on compute and transient only, has no such
dependency -- both terms are functions of the serialized byte layout, which the fee value cannot
change, so a single zero-fee probe prices the final transaction exactly.

## Reproduction

Repro branch:

https://github.com/biryukovmaxim/vprogs/tree/repro/g11-fee-probe-single-round

```bash
cargo test -p vprogs-l1-wallet --lib
```

The three tests there show the storage-mass non-convergence under the current `min_fee`:

```text
built tx pays 169000 but its own min_fee is 171900 (compute mass 1690, storage mass 1719)
built tx pays 2333400 but its own min_fee is 2431300 (compute mass 2038, storage mass 24313)
built tx pays 6500000 but its own min_fee is 6569500 (compute mass 14549, storage mass 65695)
```

They pin a real symptom, but they measure it against `min_fee` -- the wallet's own policy, which
is itself the defect -- so they are not the acceptance criterion for a fix. The transient-mass
underpayment is not covered by them at all.

## Scope

- Executed at the wallet layer only; no node or mempool was run. Both the pricing rule and the
  relay-fee constants are read from the pinned rusty-kaspa source, not observed against a live
  node.
- The `activity` (0.1 KAS UTXO) and `pay_to_address` (0.5 KAS change) shapes are reachable through
  the public builders but are not what `Wallet` currently selects; the settlement witness length
  (2000) is a stand-in.
- `RejectStorageMass` caps contextual mass at 100,000, and the settlement shape already carries
  65,695.

## Suggested fix

1. Price `min_fee` as the node does: `100 sompi/gram * max(compute_mass, normalized_transient_mass)`,
   with the transient term normalized through `block_mass_limits().raw_post().cofactors()`. Storage
   mass leaves the fee formula.
2. Treat storage mass as a funding constraint: check block fit, and either keep a storage-viable
   change output, fold an unviable one into the fee, or fund from another input -- returning a
   typed error instead of asserting when neither works.
3. If the wallet is to pay a competitive rate rather than the bare floor, run the fixpoint there,
   over the mempool's storage-inclusive priority mass.

Acceptance: each builder's output clears the node's own floor (compute and transient) for
storage-dominated, payload-dominated and witness-dominated shapes, and stays under the block-fit
storage limit.

## Merge notes

- A fix on `master` will not cover the `genesis-lock` branch's `signed_carrier_transaction`
  (`:290-300` there, which also shifts settlement to `:377-381`) -- a fourth `min_fee` site that
  exists only on that branch. Whichever branch merges second must re-apply the pricing.
- `examples/tn10-runtime` (unmerged `5d4f847`, on neither `master` nor `genesis-lock`) is
  sometimes cited as already having a fixpoint iteration. It does not: its
  `build_deposit_transaction` has the same one-pass probe and is a fifth instance.

biryukovmaxim:
Fixes #116. Addresses #103 for the activity, payout and settlement builders, which now return typed `BuildError`s instead of asserting; `signed_carrier_transaction` still panics on insufficient funding and is left to that issue.

## Build flow overview

Every builder (activity, payout, settlement) goes through the same four stages. The builder contributes only its transaction *layout*; every fee/change/input decision lives in the shared `fund` search. The sections below detail each stage; this is the map.

1. **Layout probe.** The builder supplies a `probe(n, change)` closure returning the fully *signed* transaction spending the first `n` funding candidates (plus any builder-fixed inputs, e.g. the settlement's covenant input), with the change output last when present, together with the entries all inputs spend. Signature scripts count toward mass, and the fee value never changes the byte layout, so a probe prices the final transaction exactly.
2. **Pricing.** Each probed layout gets its admission floor (`min_fee`: compute vs normalized transient, storage excluded) and, under `TargetFeerate`, the least fee reaching the estimator's rate over the node's priority mass - a fixpoint, since raising the fee shrinks the change, which raises storage mass, which raises the required fee.
3. **Prefix walk.** For `n = 1, 2, ...` over the largest-first candidate list: compute `available = inputs - fixed outputs` and the smallest storage-viable change value (`min_viable_change`). If the required fee keeps the change viable, return it with `change = available - fee` - the remainder always goes back to the caller, never into the fee. Otherwise drop the change into the fee when the layout permits it and the smaller layout's own requirement fits (bounded overpay, only when the slack is too small to legally exist as an output), else record the shortfall and walk one candidate deeper. Layouts whose change output is mandatory (activity: the change *is* the payload-carrying output) never drop, only deepen.
4. **Resolution.** On success the builder rebuilds at the chosen `(n, change)`, commits KIP-9 storage mass, and (settlement) restores the covenant witness and re-finalizes the id. On exhaustion: under a target policy, degrade to the deepest floor-affordable funding rather than fail; otherwise return a typed `BuildError` (`InsufficientFunds`, `StorageInfeasible`) so callers log and retry instead of panicking.

## Fee pricing

`min_fee` now mirrors the node's post-Toccata admission rule exactly: `100 sompi/gram * max(compute_mass, normalized_transient_mass)`, with the normalized transient term derived from `block_mass_limits().raw_post().cofactors()` - the same value the mempool uses (`check_transaction_standard.rs:146` at the pinned rusty-kaspa rev). Storage mass leaves the fee formula entirely: the node never prices it, and pricing it was what made the old single-round probe non-convergent (#116). Both remaining terms depend only on the serialized byte layout, which the fee value cannot change, so the zero-fee probe prices the final transaction exactly.

This also fixes a previously unpriced dimension: transient mass is 2 grams per serialized byte against compute's ~1, so payload-heavy activity transactions and witness-heavy settlements were underpaying the real floor and would be rejected with `RejectInsufficientTransientFee`.

The old pricing was harmless only under the relay fee the mempool used to charge: `DEFAULT_MINIMUM_RELAY_TRANSACTION_FEE` was 1000 sompi/kg = 1 sompi/gram against this crate's 100, a ~100x cushion that swallowed any mispricing. At the pinned rev it is 100_000 sompi/kg = 100 sompi/gram, exactly `MIN_FEERATE_PER_GRAM` (`mining/src/mempool/config.rs:23`); the old value survives only as `LEGACY_MINIMUM_RELAY_TRANSACTION_FEE`, for low-priority submissions before Toccata activation, which is active on every network this targets. There is no cushion left, so a shortfall is a rejection.

## Storage mass on activity transactions

`activity_transaction` never called `commit_storage_mass`, so it shipped a committed mass of zero. Deducting a fee always leaves the single output smaller than its input, which under KIP-9 is positive storage mass, so the committed value understated the real one and the node rejects the commitment. It stayed invisible because `C * (1/v_out - 1/v_in)` rounds to zero on a large coinbase UTXO; it binds exactly as the funding UTXO gets small. This is independent of the pricing defect in #116, and is fixed here with the rest.

## Funding

Storage mass still gates validity through the block-fit limit (`block_mass_limits().raw_post().storage`), which no fee can buy past. The builders now fund through a shared `fund` loop over a largest-first candidate-UTXO list:

- keep the change output when it is storage-viable (`min_viable_change`, exact via binary search over consensus's own `calc_storage_mass`);
- drop the change into the fee when it is not (bounded overpay, roughly the ~0.02 KAS viability floor), unless the layout requires the change output (activity), in which case
- add funding inputs and re-probe, walking down the candidate list;
- return typed `BuildError`s (`InsufficientFunds`, `StorageInfeasible`) instead of asserting.

All three builders (`activity_transaction`, `pay_to_address_transaction`, `settlement_transaction`) take candidate lists now. Probes never commit storage mass - a zero-value change output would divide by zero in the storage calculation - only final builds do.

`signed_carrier_transaction` is the exception: it is repriced through the new `min_fee` and commits storage mass, but still funds from a single caller-chosen UTXO and does not go through `fund`, so it neither checks change viability nor returns an error. A change value that lands under the viability floor there still commits storage mass above the block-fit limit, or panics in the storage calculation at zero. Routing it through `fund` changes its signature and its callers, so it is left to #103.

Callers updated accordingly: `Wallet` drains spent outpoints across payload batches, `FundedSettlement.fee_outpoint` becomes `fee_outpoints: Vec<_>` (the settler's retry loop excludes all of them), the sim's fee source passes its filtered spendable list, and tn10-flow's issuer tracks in-flight UTXOs from `tx.inputs`. Funding failures that collapse to `None` at the wallet API are logged with their amounts first.

## Target-feerate pricing from the fee estimate

Design and rejected alternatives recorded in #124.

The floor gets a transaction admitted; landing it competitively is priced differently. The mempool orders transactions by fee over the normalized max of compute, transient, and *storage* mass (`FeerateTransactionKey::from_tx` -> `Mass::normalized_max`), so a fee targeting an estimator rate depends on the change value it is taken from: raising the fee shrinks the change, which raises storage mass, which raises the required fee.

`fund` therefore takes a `FeePolicy`:

- `Floor` pays exactly the admission floor - the integer-only path above, unchanged and pinned byte-identical by regression tests (`floor_policy_pays_exactly_the_mempool_floor`, `floor_policy_funding_outcome_is_pinned_for_every_shape`);
- `TargetFeerate(rate)` runs a monotone fixpoint per candidate prefix, seeded at the floor, converging on the least fee satisfying the node's own `fee as f64 / mass as f64 >= rate` comparison (`rate_fee` corrects for f64 rounding, including above 2^53 where the integer gap exceeds one sompi). A fee that would push the change below its viability floor falls through to the existing drop-change / add-input restructuring; when no candidate set reaches the target, the builder degrades to the deepest floor-affordable funding (kept or dropped change) instead of erroring, so an absurd estimate can never stall settlement.

Rejected alternatives, for the record: a separate top-up phase after floor funding (duplicates the prefix/viability walk), and binary search on the fee (unsound - the sufficiency predicate is non-monotone near small change values, where the required-fee slope `r*C*p^2/v^2` exceeds 1).

`Wallet::fee_policy()` fetches `get_fee_estimate` and targets the sub-minute `normal_buckets[0]` rate, falling back to the priority bucket on an empty vector and to `Floor` (with a warning) when the RPC fails. Only `build_activity_excluding` and `prepare_settlement_excluding` fetch it; batch payload builds, payouts, one-shot settlement, and the sim stay floor-priced explicitly.

## Tests

The three repro tests from #116 are carried on this branch and now pass unweakened, reworked into 32 regression tests: node-floor coverage per shape, block-fit storage assertions, transient-domination for payload-heavy activity and witness-heavy settlement, exact keep/drop boundary via fixed-point iteration, multi-input accumulation past a payout shortfall, insufficiency and infeasibility errors, a committed-vs-recomputed storage-mass check; and for the target policy: fixpoint exactness on storage- and byte-dominated shapes against an independent priority-mass mirror, add-input and drop-change restructuring under a target rate, deepest-affordable degrade across multiple prefixes, storage-infeasible layouts barred from degrading, exact-floor-availability and zero-value-output panic guards, and `rate_fee` f64 boundary cases.

`cargo nextest run -p vprogs-l1-wallet --lib` 32/32; sim suite 8/8; workspace check and clippy clean.

## Merge notes

- `signed_carrier_transaction` was a `genesis-lock`-only fourth `min_fee` site when this branch started. That branch has since merged, so it is on master and is repriced here (`build/carrier.rs`); only its funding is out of scope, as above.
- `examples/tn10-runtime`'s `build_deposit_transaction` (unmerged `5d4f847`) is a fifth site with the same one-pass probe, untouched here.
- Zero-valued fixed outputs are outside the builders' input contract: they panic in the storage calculation on affordable prefixes (pre-existing) - validating them at the builder-args layer is a separate decision, tracked in #125.


l1/wallet/src/lib.rs
```diff
@@ -85,28 +106,25 @@ impl<'a, C: RpcApi + ?Sized> Wallet<'a, C> {
...
+                let tx = build::activity_transaction(build::ActivityTx {
                     payload,
-                    outpoint,
-                    entry,
+                    candidates: utxos.clone(),
                     keypair: self.keypair,
                     address: &self.address,
                     subnetwork_id,
                     tx_version,
+                    fee_policy: build::FeePolicy::Floor,
```
hmoog: `build_subnet_payload_transactions` hardcodes `FeePolicy::Floor` while `build_activity_excluding` and `prepare_settlement_excluding` price from the node's fee estimate, although the commit message claims activity txs are estimate-priced. If bulk payload issuance is meant to run only on quiet networks, say so in a comment; otherwise fetch `self.fee_policy()` once before the loop and reuse it (which also avoids a per-payload RPC).

l1/wallet/src/build/pricing.rs
```diff
@@ -0,0 +1,230 @@
...
+//! reconciles a target rate against a change value that shrinks as the fee grows.
+
+use kaspa_consensus_core::{
+    config::params::Params,
+    mass::{ContextualMasses, Mass, MassCalculator, UtxoCell, calc_storage_mass},
+    tx::{Transaction, UtxoEntry},
+};
+
+/// mempool floor: a transaction's fee must be at least this many sompi per mass gram.
+pub(super) const MIN_FEERATE_PER_GRAM: u64 = 100;
```
hmoog, line 19: `MIN_FEERATE_PER_GRAM = 100` mirrors the node's default relay-fee configuration, not a consensus rule. A node started with a higher minimum relay fee would silently underprice every Floor-policy transaction into rejection. Worth one doc sentence naming the config-drift risk, and possibly a follow-up that derives the floor from `get_fee_estimate`'s lowest bucket when available.

hmoog (approved):
Looks good now

PR #127, opened 2026-07-28
snapshot-core (1/4): streaming SMT builder + read-only open + raw scan cursor + snapshot codec

biryukovmaxim:
First of four stacked branches for snapshot state-sync (bringing up a fresh L2 node whose lane is older than the L1 pruning period). Base: `master`. The building blocks the save/restore branches consume:

- **`core/smt`: feed-based `StreamingBuilder<H>`** — `new(version)` / `feed(&mut wb, id, value_hash)` / `finish(&mut wb) -> root`. Writes each node into the caller's write-batch as its subtree seals and retains only a bounded left spine (≤ tree depth), holding no borrow of the batch between calls, so a restore can commit the batch and hand a fresh one to the next feed and rebuild the tree in memory bounded by depth, not leaf count. Byte-identical to `Updater` on an empty store; `build_sorted` is now a thin wrapper, gated by the existing exhaustive oracle plus commit-between-feeds and deep-spine tests.
- **`storage/rocksdb-store`: read-only open + zero-copy scan** — `open_read_only` (a live daemon can keep the store open for writing while export reads a consistent view), a plain `RawScanCursor` yielding borrowed `&[u8]` key/value pairs, and `get_pinned` for a zero-copy point lookup. No per-record allocation on the export path.
- **`state/snapshot`: streaming byte codec** — a digest-sealed `(resource_id, value)` record stream generic over a `SnapshotFormat` framing trait (magic/version) and a `Hasher` (digest). A push-style `SnapshotWriter` takes borrowed slices; a lending `SnapshotReader` yields borrowed slices from a reused buffer. Framework-agnostic: no SMT/proof concepts (the state-root self-verification lives in restore, via the builder above). Hardened against untrusted input (bad magic/version, oversized header/value/count, truncation, trailing bytes, digest mismatch).


core/smt/src/builder.rs
```diff
@@ -0,0 +1,254 @@
...
+    ///
+    /// # Panics
+    ///
+    /// In debug builds, panics if `id` is not strictly greater than the previously fed id.
+    pub fn feed<W: WriteBatch>(&mut self, wb: &mut W, id: ResourceId, value_hash: [u8; 32]) {
+        let new_id: [u8; 32] = *id;
+
+        // Seal the accumulated left side into the left child of the split with the incoming leaf.
+        if let Some(prev) = self.prev {
+            debug_assert!(new_id > prev, "build_sorted requires strictly ascending, unique leaves");
```
hmoog, line 97: A snapshot file is explicitly untrusted input, but nothing on the read path validates record ordering in release builds: `SnapshotReader::next` yields records in whatever order the file has them (the sortedness check exists only on the *writer*, which an attacker doesn't run), and `feed`'s strictly-ascending guard is a `debug_assert!`, compiled out in release.

Concretely, with duplicate ids in release: `divergence(prev, new) == 256`, so `raise(.., 257)` becomes a silent no-op, both leaves get written to the same level-256 key (last write wins), and `merge` then writes a malformed `Internal` node at leaf depth. Out-of-order ids swap sibling slots.

No panic, no error and the documented commit-per-feed pattern persists the garbage nodes to RocksDB *before* the caller's trusted-root comparison can fail. The root check does ultimately catch the forgery, but it lives in a future caller and fires only after a possibly billions-of-leaves rebuild, and no doc tells the caller to validate order or treat the store as garbage until the root matches.

Suggestion: have `SnapshotReader::next` track the previous id and return `SnapshotError::Malformed("records not sorted")` on a non-ascending id (one 32-byte compare per record), and/or make `feed` return `Err` on ordering violations in all build profiles. The reader is the natural home since the container defines the canonical on-wire order.

state/snapshot/src/container.rs
```diff
@@ -0,0 +1,616 @@
...
+    pub fn write_record(&mut self, id: &[u8; 32], value: &[u8]) -> Result<(), SnapshotError> {
+        let value_len: u32 = value.len().try_into().map_err(|_| SnapshotError::FieldTooLarge)?;
+
+        // Sortedness check against the previous record; compiled out entirely in release builds,
+        // where the caller controls the write order from the enumeration it sorted upstream.
+        #[cfg(debug_assertions)]
+        {
+            if let Some(prev) = self.prev_id {
+                debug_assert!(
+                    prev <= *id,
```
hmoog, line 167: The writer's contract and debug guard are *non-decreasing* (`prev <= *id`), but the crate's documented restore consumer (`StreamingBuilder`) requires *strictly ascending*. An exporter that re-emits a record on retry/resume writes the same id twice, passes the writer's check in both debug and release, produces a digest-valid file and restore then panics in debug (the builder's `new_id > prev` assert) or silently corrupts the tree in release (finding 1's duplicate-id path).

The doc line "id MUST be non-decreasing across calls: the canonical on-wire order" is itself wrong relative to the consumer; the check and doc should be `prev < *id`.

storage/rocksdb-store/src/raw_scan.rs
```diff
@@ -0,0 +1,162 @@
...
+/// invalidated by the next `next()`/`seek()` call.
+pub struct RawScanCursor<'a> {
+    /// The live raw iterator over the column family.
+    iter: RawIter<'a>,
+}
+
+impl<'a> RawScanCursor<'a> {
+    /// Opens a raw iterator over `cf`, positioned at the first key.
+    fn new(db: &'a DB, cf: &'a ColumnFamily) -> Self {
+        let mut iter = db.raw_iterator_cf(cf);
```
hmoog, line 25: Five column families (`Data`, `RollbackPtr`, `SmtNode`, `SmtStale`, `ProofReceipt`, per `storage/rocksdb-store/src/config.rs:52-98`) are configured with fixed-prefix extractors.

`RawScanCursor::new` uses `db.raw_iterator_cf(cf)` with default `ReadOptions` and a scan over any prefix-extractor CF runs in prefix mode which RocksDB documents as undefined across prefix boundaries.

A similar problem was fixed in `prefix_iter` which sets `total_order_seek(true)` for empty prefixes with the comment "the fixed-prefix CFs need total-order seek to span all keys" (`storage/rocksdb-store/src/store.rs:101-104`). 

The moment memtable prefix bloom or a hash index is enabled (the `Config` trait is open), `seek()`/`next()` silently skip keys and an export streams an incomplete record set that only fails much later at the trusted-root check.

The new unit test only covers `StatePtrLatest`, the one scan-relevant CF *without* an extractor. One-line fix: `raw_iterator_cf_opt` with `total_order_seek(true)`.

zk/backend/risc0/api/src/sha256.rs
```diff
@@ -12,12 +12,29 @@ mod precompile {
...
 
     /// SHA-256 hasher dispatching to the RISC-0 SHA-256 precompile via [`risc0_zkvm::sha::Impl`].
     pub struct Sha256;
 
+    /// Incremental SHA-256 state, wrapping the rust-crypto wrapper so streamed updates still
+    /// dispatch to the precompile.
+    #[derive(Default)]
+    pub struct Sha256Incremental(RustCryptoSha256);
+
+    impl IncrementalHasher for Sha256Incremental {
```
hmoog, line 25: The in-guest incremental (and one-shot) SHA-256 inherit risc0-zkp 3.0.4's `u32` bit-length arithmetic: `rust_crypto.rs:129-131` computes `bit_len` as `8 * (pos + 64 * block_len)` entirely in `u32` before zero-extending into the 64-bit SHA-256 padding field (verified in the pinned registry sources).

Any input totaling ≥ 2^32 bits (512 MiB) wraps the padding length and produces a digest that disagrees with the host's `sha2`-backed `vprogs_core_hashing::Sha256` (which uses `u64`).

A snapshot-scale digest computed on the host and recomputed in-guest would disagree, failing ZK verification with no diagnosable cause.

This is latent today (no `incremental()` call sites in `zk/` yet), but the `IncrementalHasher` doc promises the digest of "everything fed" with no cap.

Maybe we should at least add a documented limit or an upstream check before anybody feeds larger contents through the guest.

storage/rocksdb-store/src/store.rs
```diff
@@ -46,8 +46,27 @@ impl<C: Config> RocksDbStore<C> {
         }
     }
 
+    /// Opens an existing store read-only: sees data committed at open time, not writes made
+    /// through another handle afterward. Can be opened while a separate read-write handle holds the
+    /// same store open.
+    pub fn open_read_only<P: AsRef<Path>>(path: P) -> Result<Self, rocksdb::Error> {
```
hmoog, line 52: `open_read_only` constructs the store with `CanonicalChain::default()` (tip 0).

For a read-write store the chain is subsequently restored/driven by the writer; a read-only handle has no writer and stays at the default forever.

`Tree::node` gates its canonicality filter on `snapshot.tip() > 0` (`store.rs:132`), so on a read-only handle it returns the newest version `<= max_version` **regardless of canonicality** and since key encoding sorts higher versions first, an orphaned newer version shadows the canonical one.

Rollback only repoints `StatePtrLatest` (it deletes no `SmtNode` data) and pruning is an async background worker, so reorged-out state can rest on disk across restarts exactly when a read-only inspector/exporter opens the DB. 

Store::canonical_chain_manager()` would work on the RO handle, but neither the doc ("sees data committed at open time") nor the API requires it. Either restore the chain on read-only open, or document that versioned SMT reads are out of contract on a read-only handle.

Related: the read-only handle is the same type as the read-write one, so `write_batch()`/`commit()` remain callable and panic at runtime. Maybe we should add a doc note ("writes through this handle panic") to make this more obvious.

state/snapshot/src/container.rs
```diff
@@ -0,0 +1,616 @@
...
+    }
+
+    /// Appends one record (`id[32] | value_len:u32 | value`), folding it into the running digest.
+    ///
+    /// `id` MUST be non-decreasing across calls: the canonical on-wire order.
+    ///
+    /// Returns [`SnapshotError::FieldTooLarge`] if `value` is longer than `u32::MAX`, rather than
+    /// silently truncating the on-wire length prefix.
+    pub fn write_record(&mut self, id: &[u8; 32], value: &[u8]) -> Result<(), SnapshotError> {
+        let value_len: u32 = value.len().try_into().map_err(|_| SnapshotError::FieldTooLarge)?;
```
hmoog, line 159: `write_record` accepts any value up to `u32::MAX` while the paired reader rejects `value_len > MAX_VALUE_LEN` (256 MiB); likewise `open` accepts headers over `MAX_HEADER_LEN`.

A node can export a snapshot containing one 300 MiB resource value: the export completes, the digest verifies clean, and every later restore fails with `Malformed`, discovered only at restore time, when the snapshot may be the sole surviving copy of the state.

We should most probably enforce the caps symmetrically at write time.

state/snapshot/src/container.rs
```diff
@@ -0,0 +1,616 @@
...
+        self.hw.write_all(&value_len.to_le_bytes())?;
+        self.hw.write_all(value)?;
+        self.written += 1;
+        Ok(())
+    }
+
+    /// Writes the trailing digest over every byte written since [`open`](Self::open).
+    pub fn finish(self) -> Result<(), SnapshotError> {
+        // Debug-only: the number of records written must match the count declared at open.
+        debug_assert_eq!(
```
hmoog, line 185: `finish` only `debug_assert_eq!(written, expected)`.

A release-mode producer bug (e.g. the count pass and the stream pass observing different implicit snapshots) ships a snapshot with a *valid* digest that `finish()` reports as success; the mismatch surfaces only at restore, as a confusing `Truncated`/`DigestMismatch` misdiagnosed as file corruption possibly after the source state is gone.

`finish` already returns `Result`, so a real `Malformed` error in release is free.

state/snapshot/src/container.rs
```diff
@@ -0,0 +1,616 @@
...
+}
+
+/// Fills `buf` completely and folds the bytes read into `hasher`, mapping any short read to
+/// [`SnapshotError::Truncated`].
+fn read_exact_fold<R: Read, Inc: IncrementalHasher>(
+    r: &mut R,
+    buf: &mut [u8],
+    hasher: &mut Inc,
+) -> Result<(), SnapshotError> {
+    r.read_exact(buf).map_err(|_| SnapshotError::Truncated)?;
```
hmoog, line 344: Maps *every* `read_exact` error to `SnapshotError::Truncated`, making the `Io` variant unreachable on those paths while the value-read path (`read_to_end`) surfaces the same faults as `Io`, so diagnosis flips depending on which field the fault lands in.

An operator restoring from a failing disk sees "snapshot truncated" and re-downloads a good snapshot instead of diagnosing the disk.

Imho, only `ErrorKind::UnexpectedEof` should become `Truncated`:

```rust
r.read_exact(buf).map_err(|e| match e.kind() {
    std::io::ErrorKind::UnexpectedEof => SnapshotError::Truncated,
    _ => SnapshotError::Io(e),
})?;
```

state/snapshot/src/container.rs
```diff
@@ -0,0 +1,616 @@
...
+    /// the whole snapshot. Also rejects (`Malformed`) any bytes left in the reader once the
+    /// digest has been consumed, so a file with a correct digest but junk appended after it does
+    /// not "verify".
+    pub fn finish(self) -> Result<(), SnapshotError> {
+        if self.remaining != 0 {
+            return Err(SnapshotError::Malformed("finish called before all records were read"));
+        }
+        let mut reader = self.reader;
+        let mut digest = [0u8; 32];
+        reader.read_exact(&mut digest).map_err(|_| SnapshotError::Truncated)?;
```
hmoog, line 324: Maps *every* `read_exact` error to `SnapshotError::Truncated`, making the `Io` variant unreachable on those paths while the value-read path (`read_to_end`) surfaces the same faults as `Io`, so diagnosis flips depending on which field the fault lands in.

An operator restoring from a failing disk sees "snapshot truncated" and re-downloads a good snapshot instead of diagnosing the disk.

Imho, only `ErrorKind::UnexpectedEof` should become `Truncated`:

```rust
reader.read_exact(buf).map_err(|e| match e.kind() {
    std::io::ErrorKind::UnexpectedEof => SnapshotError::Truncated,
    _ => SnapshotError::Io(e),
})?;
```

storage/rocksdb-store/src/raw_scan.rs
```diff
@@ -0,0 +1,162 @@
...
+    /// directly from RocksDB for zero-copy forward scans (e.g. streaming a resource-id-sorted CF
+    /// like `latest_ptr`).
+    pub fn raw_scan(&self, ns: StateSpace) -> RawScanCursor<'_> {
+        RawScanCursor::new(&self.db, self.cf(&ns))
+    }
+
+    /// Zero-copy point lookup: the value bytes for `key` in `ns`, borrowed straight from RocksDB
+    /// with no heap copy, or `None` if absent. Panics on a rocksdb error, matching `get`.
+    pub fn get_pinned(&self, ns: StateSpace, key: &[u8]) -> Option<DBPinnableSlice<'_>> {
+        match self.db.get_pinned_cf(self.cf(&ns), key) {
```
hmoog, line 77: The module doc and the container doc both advertise the cursor-key + `get_pinned`-value pairing, but the cursor pins a consistent view while `get_pinned` reads the live DB.

No snapshot-scoped point read exists in the store API. An exporter walking a `latest_ptr` cursor pinned at T0 and fetching values via `get_pinned` from another CF can, under a concurrent commit/prune, get `None` or newer bytes, mixing state from two points in time.

Likewise, `SnapshotWriter::open`'s required `record_count` forces a counting cursor whose implicit snapshot differs from the streaming cursor's.

We likely either need a snapshot-scoped point read or an explicit export contract requiring a quiesced/read-only store.

storage/rocksdb-store/src/raw_scan.rs
```diff
@@ -0,0 +1,162 @@
...
+impl<'a> RawScanCursor<'a> {
+    /// Opens a raw iterator over `cf`, positioned at the first key.
+    fn new(db: &'a DB, cf: &'a ColumnFamily) -> Self {
+        let mut iter = db.raw_iterator_cf(cf);
+        iter.seek_to_first();
+        Self { iter }
+    }
+
+    /// Whether the cursor is positioned at a live key (not exhausted or errored).
+    pub fn valid(&self) -> bool {
```
hmoog, line 31: `valid()` returns false for both a finished scan and a mid-scan RocksDB error, and nothing forces callers to consult `status()`.

The natural `while cursor.valid()` export loop silently truncates on block corruption or an I/O error and combined with finding at `state/snapshot/src/container.rs:185` (release-mode no-op count check), that produces a *short snapshot with a valid digest* reported as a successful export.

This also breaks the store's otherwise uniform panic-on-error convention (`get`, `prefix_iter`, `get_pinned` all panic loudly).

Maybe we could consider checking `status()` inside `next()`/`valid()` and panicking, matching the rest of the store.

core/smt/tests/builder_oracle.rs
```diff
@@ -0,0 +1,540 @@
...
+        1,
+        "mixed-shortcut-and-deep-pair",
+    );
+}
+
+// -- Contract-violation guards (debug builds) --
+
+/// A descending stream violates the ascending contract and must panic in debug builds.
+#[test]
+#[should_panic(expected = "strictly ascending")]
```
hmoog, line 229: `#[should_panic]` oracle tests fail under `cargo test --release` for me.

It's backed by a `debug_assert` but lack the `#[cfg_attr(not(debug_assertions), ignore = ...)]` guard that the analogous container test (`container.rs:431`) carries.

In a release test run the assert is compiled out, the streams build without panicking, and the test fails with "test did not panic as expected" on a healthy tree.

core/smt/tests/builder_oracle.rs
```diff
@@ -0,0 +1,540 @@
...
+    build_sorted::<_, Sha256>(
+        &mut cap,
+        1,
+        leaves.into_iter().map(|(id, value_hash)| Leaf { id, value_hash }),
+    );
+}
+
+/// Duplicate ids violate the uniqueness contract and must panic in debug builds.
+#[test]
+#[should_panic(expected = "strictly ascending")]
```
hmoog, line 243: `#[should_panic]` oracle tests fail under `cargo test --release` for me.

It's backed by a `debug_assert` but lack the `#[cfg_attr(not(debug_assertions), ignore = ...)]` guard that the analogous container test (`container.rs:431`) carries.

In a release test run the assert is compiled out, the streams build without panicking, and the test fails with "test did not panic as expected" on a healthy tree.

core/smt/tests/builder_oracle.rs
```diff
@@ -0,0 +1,540 @@
...
+
+/// Records every `put_node` write, keyed by the 34-byte node key encoding.
+#[derive(Default)]
+struct Capture {
+    nodes: BTreeMap<[u8; 34], (u64, Vec<u8>)>,
+}
+
+impl WriteBatch for Capture {
+    fn put_node(&mut self, key: &Key, version: u64, data: &Node) {
+        self.nodes.insert(key.encode(), (version, data.encode()));
```
hmoog, line 56: `Capture::put_node` uses `BTreeMap::insert` and drops the returned prior entry, so the oracle compares last-write-wins end states rather than the actual sequence of writes the module doc guarantees ("the full set of node writes"); the RocksDB variants also only read end state.

A builder regression that emits an extra `put_node` with wrong bytes to a key later overwritten by the correct write (exactly the shape the release-mode duplicate-id path produces at level 256) is invisible to every assertion.

Fix: assert the insert returns `None` (an empty-store build must never write the same key twice).

node/test-utils/src/l1_node.rs
```diff
@@ -203,14 +203,37 @@ impl L1Node {
...
+            .await;
+
+        // The node's UTXO index is fed by consensus notifications on its own task, so submit_block
+        // returning says nothing about what a UTXO query sees. Poll until the mined coinbases are
+        // visible; otherwise a wallet build issued right after mining can find no candidates.
+        let deadline = tokio::time::Instant::now() + Duration::from_secs(10);
+        loop {
+            let spendable =
+                self.wallet().fetch_spendable_utxos().await.expect("fetch spendable utxos");
+            if spendable.len() >= num_utxos {
```
hmoog, line 225: The predicate `spendable.len() >= num_utxos` counts every UTXO in the wallet (`fetch_spendable_utxos` maturity-filters only coinbases), so on a wallet already holding change from earlier spends the poll can succeed on stale entries before the newly mined coinbases are indexed, reintroducing the index-lag race this loop was added to close.

Latent (every current caller polls a fresh wallet once), but a baseline-delta or coinbase-only count would close it for the first test that tops up mid-test.

node/framework/tests/e2e.rs
```diff
@@ -37,18 +37,25 @@ fn create_node(l1: &L1Node, temp_dir: &TempDir) -> Node<RocksDbStore, TestNodeVm
...
-            .await;
-        tx_hashes.push(txs[0].id());
-        l1.mine_block(&txs).await;
+            })
+        })
+        .collect();
+    // Build every payload transaction from one read of the wallet's spendable set. The builder
+    // subtracts what each transaction spends, so no read has to wait on the node's UTXO index
+    // catching up with the block mined for the previous payload.
+    let txs = l1.build_payload_transactions(payloads).await;
```
hmoog, line 53: The single up-front `build_payload_transactions(payloads)` draws on a pool of exactly `count` UTXOs (callers pass `count == mine_utxos(count)`) with subtract-only accounting and no replenishment.

Any transaction that consumes 2+ candidate UTXOs (fee-estimate change, small non-coinbase UTXOs entering the pool) starves a later payload and panics in the wallet builder.

The removed per-iteration re-fetch was the headroom.

Latent today but mining `count + 1` UTXOs or asserting the invariant would restore the margin.

state/snapshot/src/container.rs
```diff
@@ -0,0 +1,616 @@
...
+        if self.hasher.finalize() != digest {
+            return Err(SnapshotError::DigestMismatch);
+        }
+
+        let mut probe = [0u8; 1];
+        if reader.read(&mut probe)? != 0 {
```
hmoog, line 330: The trailing-byte probe in `SnapshotReader::finish` is a single raw `read()`, so `ErrorKind::Interrupted` fails verification of a valid snapshot (every other read path retries EINTR via `read_exact`/`read_to_end`). 

Maybe loop on `Interrupted`.

core/smt/src/builder.rs
```diff
@@ -0,0 +1,254 @@
...
+
+    /// Feeds the next leaf into the stream.
+    ///
+    /// Ids must arrive strictly ascending and unique. The builder holds no borrow of `wb` after
+    /// this call returns, so the caller may commit `wb` and pass a fresh batch to the next `feed`.
+    ///
+    /// # Panics
+    ///
+    /// In debug builds, panics if `id` is not strictly greater than the previously fed id.
+    pub fn feed<W: WriteBatch>(&mut self, wb: &mut W, id: ResourceId, value_hash: [u8; 32]) {
```
hmoog, line 92: The contract requires every `value_hash != EMPTY_HASH` but nothing checks it, even in debug; a violating feed silently diverges from the `Updater` oracle (which treats `EMPTY_HASH` as deletion).

A `debug_assert!` would make the contract self-enforcing in tests.

core/smt/tests/builder_oracle.rs
```diff
@@ -0,0 +1,540 @@
...
+        &mut cap,
+        0,
+        leaves.into_iter().map(|(id, value_hash)| Leaf { id, value_hash }),
+    );
+}
+
+// -- RocksDB integration --
+
+/// The left child key of `key` (bit 0 at the current level).
+fn left_child(key: &Key) -> Key {
```
hmoog, line 271: The test-local `left_child`/`right_child` helpers re-implement the existing `pub(crate)` `Key::left_child`/`right_child` (`core/smt/src/key.rs:33-46`).

PR #128, opened 2026-07-28
snapshot-save (2/4): vprun snapshot save (read-only export at latest settlement)

biryukovmaxim:
Second of four stacked branches for snapshot state-sync. Base: `snapshot-core`.

`vprun snapshot save`: open the local node's RocksDB read-only (a live daemon can keep writing) and export the L2 state as of the latest retained settlement to a self-verifying file.

- Anchors on the settlement's `block_prove_to` (the last block of the proven bundle, whose SMT root the settlement attests to), NOT the later block the settlement transaction landed in; the store's own authenticated root at that batch is checked against the on-chain settlement root before writing.
- Bounded-memory, allocation-free enumeration: a small corrections map holds only the post-settlement churn (via rollback pointers); two passes over `latest_ptr` use the zero-copy `RawScanCursor` (borrowed 32-byte ids) and `get_pinned` (borrowed values) to stream records straight into the push-style `SnapshotWriter`, so nothing is allocated per resource.
- Carries a typed header (identity + the settlement point) so a restored node resumes from `block_prove_to` and adopts the covenant without referencing the pruned deploy.


runner/src/snapshot/save.rs
```diff
@@ -0,0 +1,496 @@
...
+        });
+    }
+
+    // Bounded-memory enumeration: `corrections` holds only resources changed after N_S (the churn
+    // of a few post-settlement batches, never the full resource set); a resource absent here kept
+    // its current latest version back to N_S. `pinned` keeps each resource's first post-S touch,
+    // since that rollback entry is the version held right before that batch, i.e. its value at N_S.
+    let mut corrections: HashMap<ResourceId, u64> = HashMap::new();
+    let mut pinned: HashSet<ResourceId> = HashSet::new();
+    for idx in (n_s + 1)..=tip.index() {
```
hmoog: The loop `for idx in (n_s + 1)..=tip.index()` iterates every id in the range, canonical or not. Rollback entries of orphaned batches are retained on disk until pruning (`scheduling/scheduler/src/rollback.rs` only repoints `StatePtrLatest`; deletes happen in `pruning_worker.rs`), so they are read here as if they were canonical history. Both the scheduler's rollback (`rollback.rs:56`) and the pruning worker gate the same iteration on `is_canonical(index)`; the save routine does not.

For orphans that forked at or above `n_s` this happens to be harmless (the lowest-id touch always sees the version at `n_s`). It breaks when `n_s`'s own block was orphaned and later returned via the manager's returning-block branch (`storage/canonical-chain/src/manager.rs:47-61`, which re-canonicalizes the existing id):

```
1 canonical
2 = block B, parent 1, r1: v1 -> v2, rollback (2, r1) = 1
   reorg to 1
3 = block C, parent 1, r1: v1 -> v3, rollback (3, r1) = 1     (orphaned later)
   reorg back to 1; B returns and keeps id 2; r1 latest = 2
4 = block D, parent 2
settlement proves to B: n_s = 2, root(2) matches, check passes
corrections over 3..=4: idx 3 pins r1 -> 1
```

The snapshot emits `r1` at version 1 (`v1`) while the state at `n_s` has `v2`. Digest valid, root check passed, records wrong. Fix: skip `idx` where `_canonical.chain().snapshot().is_canonical(idx)` is false (the handle from line 87 already has the replayed view). Add a test with exactly this geometry; the existing reorg test (`save.rs:431-495`) has an empty corrections range (`n_s == tip`), so the orphan path is untested.

runner/src/snapshot/save.rs
```diff
@@ -0,0 +1,496 @@
...
+        for (rid_bytes, old_version) in StatePtrRollback::iter_batch(&store, idx) {
+            let rid: ResourceId =
+                borsh::from_slice(&rid_bytes).expect("corrupted rollback resource id");
+            if pinned.insert(rid) {
+                corrections.insert(rid, old_version);
+            }
+        }
+    }
+
+    // Write the snapshot in a single streaming pass. The `store.root(n_s) == settlement.new_state`
```
hmoog: The comment says the `store.root(n_s) == settlement.new_state` check "is the same authenticated SMT root the records here are read back from", but it is not: the root comes from the `SmtNode` CF, the records from `StatePtrLatest` + `StatePtrRollback` + `StateVersion`.

Any divergence between the two paths (see comment above, a missing version entry, a future pruning change) yields a digest-valid snapshot that fails only at restore, possibly after the source node is gone. 

Commit 7 removed the in-memory rebuild as "redundant"; `StreamingBuilder` with a discard `WriteBatch` (4 no-op methods, `core/smt/src/write_batch.rs`) recomputes the root over the emitted stream with memory bounded by tree depth and one hash per record. For a one-shot export tool that cost is worth a save-time `RootMismatch` instead of a restore-time surprise.

If we decide against it, we should at least fix the comment so it stops claiming the records are covered.

runner/src/snapshot/save.rs
```diff
@@ -0,0 +1,496 @@
...
+    // the order `SnapshotWriter::write_record` requires, so no sort is needed. A resource is
+    // emitted at its version as of N_S: `corrections` overrides the current latest for resources
+    // touched after N_S, and a resource CREATED after N_S surfaces as a correction of exactly 0 and
+    // is skipped, since no version existed at N_S. A resource whose value at N_S is empty (or has
+    // no stored version) is skipped too: the live store keeps an empty value to mark a deletion
+    // and shadow the prior version, but a restore rebuilds a fresh tree where empty and absent
+    // coincide and the SMT already treats empty as absent, so the record would only bloat the
+    // file.
+    let mut emit_cursor = store.raw_scan(StateSpace::StatePtrLatest);
+    while emit_cursor.valid() {
```
hmoog: `while emit_cursor.valid()` exits on both exhaustion and error (`RawScanCursor::valid`), and `status()` is never consulted. With the backpatched count, a scan that dies halfway writes the true (short) count, a matching digest, and returns `Ok`.

Check `emit_cursor.status()` after the loop and return an error. (Same class as the thing I commented on in the `snapshot-core` review, but this is the first caller and now the file can no longer betray the truncation.)

runner/src/snapshot/save.rs
```diff
@@ -0,0 +1,496 @@
...
+
+        let ver = corrections.get(&ResourceId::from(id)).copied().unwrap_or(latest);
+        if ver != 0 {
+            // StateVersion key layout: version (u64 BE) || resource_id (borsh); `ResourceId`'s
+            // borsh encoding of its single `[u8; 32]` field is the field's bytes verbatim, so `id`
+            // doubles as the borsh-encoded key suffix with no re-encode.
+            let mut version_key = [0u8; 40];
+            version_key[..8].copy_from_slice(&ver.to_be_bytes());
+            version_key[8..].copy_from_slice(&id);
+            let value = store.get_pinned(StateSpace::StateVersion, &version_key);
```
hmoog: `value.as_deref().filter(|v| !v.is_empty())` collapses `None` (no version entry for a version the latest/rollback pointers say exists) into the "empty, skip" branch. Within a read-only view a version referenced from `>= root` cannot have been pruned, so `None` here is corruption or a reconstruction bug and should be a hard `SaveError`, not a dropped record (we only prune dropped data later).

runner/src/snapshot/save.rs
```diff
@@ -0,0 +1,496 @@
...
+    // streams the records, then backpatches the true count, so nothing here ever holds the full
+    // resource set.
+    let header = SnapshotHeader {
+        covenant_id,
+        lane_id,
+        bootstrap_txid,
+        committed_index: n_s,
+        chain_block_metadata: meta_at_s,
+    };
+    let mut file = std::fs::File::create(out)?;
```
hmoog: `File::create(out)` truncates any existing snapshot before the routine knows it will succeed; an I/O error or a kill mid-stream leaves a partial file at the operator's chosen path with no marker.

We should write to a sibling temp path and `rename` on success, remove on failure. Separately, `File` is unbuffered and `write_record` issues three `write_all` calls per record, so this is three syscalls per resource; wrap in `BufWriter` (it implements `Seek`) and `flush` before returning.

runner/src/snapshot/save.rs
```diff
@@ -0,0 +1,496 @@
...
+    // A fresh handle's canonical-chain oracle is empty, which reads every stored node version as
+    // canonical. Replay the persisted batch log into it, as a live node does at start-up, so the
+    // root check below steps over versions a reorg orphaned.
+    let _canonical = store.canonical_chain_manager::<ChainBlockMetadata>();
+
+    // Identity comes from the JSON file, not the DB.
+    let identity = PersistedState::load(data_dir);
+    let covenant_id = identity.covenant_hash().ok_or(SaveError::NoIdentity)?;
+    let lane_id = identity.lane_id.ok_or(SaveError::NoIdentity)?;
+    let bootstrap_txid = identity.bootstrap_txid().unwrap_or_else(|| Hash::from_bytes([0u8; 32]));
```
hmoog: `unwrap_or_else(|| Hash::from_bytes([0u8; 32]))` bakes a bogus covenant outpoint `(0x00.., 0)` into a snapshot that otherwise verifies. A restored node's settler would fail much later with no pointer back to the cause. Treat it like `covenant_id`/`lane_id` (return `NoIdentity`), or make the header field an `Option`.

state/ptr-latest/src/lib.rs
```diff
@@ -38,4 +38,43 @@ impl StatePtrLatest {
         let key = borsh::to_vec(resource_id).expect("failed to serialize ResourceId");
         wb.delete(StateSpace::StatePtrLatest, &key);
     }
+
+    /// Enumerate every (resource_id, current_version) pair. Full scan of the latest-ptr CF.
+    pub fn iter_all<S>(store: &S) -> impl Iterator<Item = (ResourceId, u64)> + '_
```
hmoog: Commit 1 added these; commit 8 moved the save path to `raw_scan`, and nothing else calls it (only `iter_all`'s own test remains).

state/ptr-latest/Cargo.toml
```diff
@@ -7,3 +7,7 @@ version.workspace = true
 borsh.workspace                = true
 vprogs-core-types.workspace    = true
 vprogs-storage-types.workspace = true
+
```
hmoog: This was only added to implement iter_all on StatePtrLatest, which seems to no longer be used (see the other comment in `state/ptr-latest/src/lib.rs:43`

storage/rocksdb-store/src/store.rs
```diff
@@ -118,6 +118,12 @@ impl<C: Config> Store for RocksDbStore<C> {
     fn canonical_chain(&self) -> CanonicalChain {
         self.canonical.clone()
     }
+
+    fn scan(&self, state_space: StateSpace) -> PrefixIterator<'_> {
```
hmoog: Commit 1 added this; commit 8 moved the save path to `raw_scan`, and nothing else calls either (only `iter_all`'s own test remains).

`scan` is byte-for-byte `prefix_iter(ns, &[])` except it omits `total_order_seek(true)`, so on the five prefix-extractor CFs it is the wrong full-scan primitive.

storage/types/src/store.rs
```diff
@@ -41,4 +41,8 @@ pub trait Store: Tree + Clone + Send + Sync + 'static {
         // Replay them into a manager over this store's oracle.
         CanonicalChainManager::new(self.canonical_chain(), entries)
     }
+
+    /// Iterate over every key-value pair in the given state space, in key order.
+    /// Unlike `prefix_iter`, this never bounds on a prefix; use it to enumerate a whole CF.
+    fn scan(&self, state_space: StateSpace) -> PrefixIterator<'_>;
```
hmoog: Commit 1 added this; commit 8 moved the save path to `raw_scan`, and nothing else calls either (only `iter_all`'s own test remains).

`scan` is byte-for-byte `prefix_iter(ns, &[])` except it omits `total_order_seek(true)`, so on the five prefix-extractor CFs it is the wrong full-scan primitive.

runner/Cargo.toml
```diff
@@ -26,14 +26,23 @@ test-utils = ["vprogs-zk-backend-risc0-settler/test-utils"]
 # vprogs internal: the node/bridge/scheduler wiring, the proving stack, covenant bootstrap, and the
 # ELF loaders + dev-mode probe are reused rather than reimplemented.
 vprogs-core-atomics                = { workspace = true }
+vprogs-core-hashing                = { workspace = true }
```
hmoog: only used from `#[cfg(test)]` code in `save.rs` and belong in `[dev-dependencies]`

runner/Cargo.toml
```diff
@@ -26,14 +26,23 @@ test-utils = ["vprogs-zk-backend-risc0-settler/test-utils"]
...
 vprogs-core-smt                    = { workspace = true }
+vprogs-core-types                  = { workspace = true }
 vprogs-l1-bridge                   = { workspace = true }
 vprogs-l1-types                    = { workspace = true }
 vprogs-l1-wallet                   = { workspace = true }
 vprogs-node-framework              = { workspace = true }
 vprogs-scheduling-scheduler        = { workspace = true }
+vprogs-state-batch-metadata        = { workspace = true }
+vprogs-state-metadata              = { workspace = true }
+vprogs-state-ptr-latest            = { workspace = true }
```
hmoog: only used from `#[cfg(test)]` code in `save.rs` and belong in `[dev-dependencies]`

runner/Cargo.toml
```diff
@@ -26,14 +26,23 @@ test-utils = ["vprogs-zk-backend-risc0-settler/test-utils"]
...
+vprogs-state-ptr-rollback          = { workspace = true }
+vprogs-state-snapshot              = { path = "../state/snapshot" }
+vprogs-state-version               = { workspace = true }
```
hmoog: only used from `#[cfg(test)]` code in `save.rs` and belong in `[dev-dependencies]`

runner/Cargo.toml
```diff
@@ -26,14 +26,23 @@ test-utils = ["vprogs-zk-backend-risc0-settler/test-utils"]
...
 vprogs-l1-bridge                   = { workspace = true }
 vprogs-l1-types                    = { workspace = true }
 vprogs-l1-wallet                   = { workspace = true }
 vprogs-node-framework              = { workspace = true }
 vprogs-scheduling-scheduler        = { workspace = true }
+vprogs-state-batch-metadata        = { workspace = true }
+vprogs-state-metadata              = { workspace = true }
+vprogs-state-ptr-latest            = { workspace = true }
+vprogs-state-ptr-rollback          = { workspace = true }
+vprogs-state-snapshot              = { path = "../state/snapshot" }
```
hmoog: the only internal dep not using `workspace = true`

runner/src/snapshot/save.rs
```diff
@@ -0,0 +1,496 @@
...
+    let root_cp: Checkpoint<ChainBlockMetadata> = StateMetadata::root(&store);
+    let settlement: SettlementInfo =
+        tip.metadata().last_settlement.ok_or(SaveError::NoSettlement)?;
+
+    // Find N_S: the batch whose block is settlement.block_prove_to, NOT
+    // settlement.containing_block (the later block the settlement transaction landed in).
+    // store.root(n_s) is the state root the settlement actually attests to. Walk down from the tip.
+    let mut n_s = tip.index();
+    let mut meta_at_s: ChainBlockMetadata;
+    loop {
```
hmoog: The loop decodes `BatchMetadata` for every id from the tip down to the match (each a borsh decode the canonical-chain replay on line 87 already performed) and does not check canonicality of the id it lands on.

`CanonicalChainManager::id(&block_hash)` (`manager.rs:87`) and `metadata(id)` (`manager.rs:92`) should give the same answer in O(1); combine with `is_canonical(id)` so a settlement pointing at an orphaned block is reported as `SettlementNotRetained`/a dedicated error instead of falling through to `RootMismatch`.

Issue #87, opened 2026-06-23
Migrate to never-reused canonical batch ids (retention, not truncation)


hmoog:
## Summary

Replace the `VirtualChain`'s truncate-and-reuse batch index with the canonical chain's
**never-reused, monotonic batch ids** plus a lock-free `is_canonical(id)` oracle, and make a reorg a
**re-selection** (flip canonical bits, retain orphan data) instead of destructive truncation. The
end state: one id space (canonical id = batch index = storage key), reorgs that don't delete state,
and `VirtualChain` removed.

This is staged so each step is an independent, mergeable PR. The hard, coupled core is isolated to
one PR; everything around it lands before or after on its own.

## Background

There are two id spaces today and they diverge after a reorg:

- **`VirtualChain` index** (`l1/bridge/src/virtual_chain.rs`): a dense counter that is reset
  backwards and **reused** on rollback. It is the `Checkpoint.index` that the scheduler keys
  `BatchMetadata`, `StateVersion`, SMT versions, etc. on.
- **`CanonicalWriter` id** (`storage/canonical-chain`): monotonic, **never reused** across reorgs,
  with an `is_canonical(id)` bitmask oracle. Currently wired in parallel but unused.

A reorg today is destructive truncation (`scheduling/scheduler/src/rollback.rs`): it deletes
`StateVersion` data, `StatePtrRollback`, `BatchMetadata`, the `CanonicalChain` CF, and SMT nodes for
the rolled-back range, and reuses the freed indexes. The goal model retains orphan branches so a
re-reorg is re-selection, not re-execution.

## Target model

- Batch ids are monotonic and **never reused**; a reorg appends the heavier branch above the
  orphaned ids (so the highest id is always canonical).
- A lock-free `is_canonical(id)` oracle answers canonical-ness; a reorg flips bits, it does not
  delete.
- Orphan state is **retained** and reclaimed only at finality (pruning).
- One id space: canonical id = batch index = storage key.

## Key design decisions

- **State reads stay point queries (repoint, not oracle-on-reads).** `StatePtrLatest` becomes the
  "latest *canonical* version" pointer, repointed (not deleted) on reorg via the existing
  `StatePtrRollback` mechanism. State reads never consult the oracle.
- **The SMT is the one oracle consumer.** It is version-addressed with no pointer to repoint, so
  `Tree::node` must resolve "highest version `<= tip` whose id `is_canonical`" against a pinned
[...]

Issue #66, opened 2026-06-01
Persistent storage for proven receipts


hmoog:
## Summary

Receipts are currently held only in memory. After a process restart or fast-forward, batches that were already proven must be re-proven from scratch even though their receipts existed. Add a persistent store keyed by `(image_id, claim_digest)` so the prover can resume work without throwing away completed proofs.

## Motivation

- **Restart resilience**: a restarted prover picks up from cached receipts instead of redoing everything.
- **Fast-forward**: when another prover settles ahead of us, we match our cache against their tip and prove only the suffix.
- **Reorg recovery**: combined with the reorg orchestrator (separate issue), the cache lets us discard only the invalidated batches and keep the rest.

## Design

A receipt is uniquely identified by `(image_id, claim_digest)`:

- `image_id` distinguishes circuits (tx-processor, batch-processor, batch-aggregator).
- `claim_digest` is the hash of the receipt's claim (journal plus public outputs); two receipts with the same `(image_id, claim_digest)` are interchangeable.

```rust
trait ProofStore {
    fn get(&self, image_id: &[u8; 32], claim_digest: &[u8; 32]) -> Option<Receipt>;
    fn put(&self, image_id: &[u8; 32], claim_digest: &[u8; 32], receipt: &Receipt);
    fn delete(&self, image_id: &[u8; 32], claim_digest: &[u8; 32]);
}
```

Storage backend: RocksDB column family fits the existing storage stack. Key layout: `image_id || claim_digest`; value: serialized receipt.

The prover loop checks the cache by claim before invoking the prover. On successful proof generation, the receipt is written. The reorg orchestrator (separate issue) calls `delete` to invalidate.

## Scope

- New trait + RocksDB-backed impl, probably in `zk/storage/...` or alongside the existing storage stack.
- Integrate into the prover driver: cache check before generation, write on completion.
- Expose an invalidation hook for the reorg orchestrator.
- Tests: store-and-retrieve round-trip; restart simulation; invalidation correctness.

## Acceptance

- A simulated prover restart resumes from previously-completed batch proofs without redoing them.
- An invalidation call removes the targeted entries and leaves others untouched.
- Storage layout supports the access patterns expected by the orchestrator (point lookup, point delete).

## Out of scope

- The reorg orchestrator that decides what to invalidate (separate issue).
- Cross-prover proof sharing or an on-chain proof registry (future concern).
- Garbage collection of orphaned entries (can be a follow-up; for now the orchestrator handles all deletions explicitly).

PR #148, opened 2026-09-20
settler, aggregate-prover: resume unsettled bundles after a restart

biryukovmaxim:
## The incidents

Two tn10 production wedges with the same signature: the settler submits a settlement, the tx is mined and accepted within ~1 s, but the settlement watch never advances and the single-flight settler waits forever ("unconfirmed after Ns"). Nothing settles afterward while execution and proving continue; the node looks healthy from the L2 side. After a prover restart the lane stayed idle (TN5). Root cause (investigated on the live systems): the bridge's Full-verbosity `get_virtual_chain_from_block_v2` catch-up fetch over a dust-storm window exceeded the RPC client's 60 s request timeout and entered a retry livelock, freezing the bridge tip below the settlement's accepting block; the notification-only confirm wait turned that transport stall into a permanent settler wedge. The same 60 s livelock is the long-known slow-link catch-up defect; the transport fix (chunked/reduced-verbosity catch-up) is separate and NOT part of this branch.

## What this branch does

Settlements resume after a prover restart, receipts are reused rather than re-proved (only the cheap aggregate re-runs), a competitor that settles during downtime is handled without forking, and the settler can no longer wedge on a mined-but-unobserved settlement.

- Settlement journal (`state/settlement-journal`, new): proved-but-unsettled bundles are journaled at artifact publication (`end_index`, `from_block`, `block_prove_to`, `seq_commit`) over a new `StateSpace::SettlementJournal` column family; entries compact away against the covenant tip.
- Startup resume: unsettled journal entries are re-loaded from the persisted aggregate receipts and re-fed to the settler ahead of new work; settled prefixes compact; a straddled entry is split at the competitor boundary and its suffix re-aggregated from cached per-batch receipts. The resume gate never parks: with no settlement observation available yet it re-feeds the tail unchanged and lets proving proceed.
- Multi-pass resume: a settlement advance re-runs the resume pass scoped to the pre-restart tail, so a competitor settlement that lands during downtime splits the straddler once the bridge observes it, not only at startup.
- Committed-gap coverage: batches committed but never bundled before the kill (the TN5 restarted-idle signature) are re-formed into a bundle at startup from cached per-batch receipts, bounded at the contiguously-durable prefix.
- Chain-derived settled-ness: before every submission the settler probes the covenant outpoint's on-chain liveness; a spent outpoint is superseded without submitting, so the chain may already hold the tx we would re-submit and we adapt instead of forking.
- Confirm-wait backstop: the warn tick already polls the covenant address UTXOs; a spent outpoint now resolves the confirm wait from the submission's own data instead of deferring to a watch that may be frozen (Landed requires our own continuation UTXO live on chain, distinguishing our landing from a competitor's Superseded). The settler never waits forever on a tx the chain already holds. Riding along: the old resubmit-on-drop probe matched a typed not-found error variant the wRPC client never delivers, so it had been inert against real nodes; the probe now accepts the actual error shape.
- Bridge observability: one INFO line per settlement tx at chain-block ingestion (txid + block + daa), so a future watch freeze is visible at ingestion in minutes.
- Durability premise: builds on the write-worker shutdown drain from the base of this stack (#147) — a write accepted before shutdown commits and opens its confirmation latch; the restart resume's receipt and metadata durability depends on it.

## Tests

`two_provers_contend` grows from 7 to 13 integration tests (dev-mode proofs, simnet L1): warm restart settles the pending tail; competitor sweep during downtime; split at the competitor boundary with no new carriers; own settlement landing during downtime (no fork, no double submission, settles past the landed tip); committed-but-unjournaled gap coverage (red-verified against the pinned wedge); blind-watch confirm via the chain poll.

## Out of scope (open)

- Fee-bump and resubmit-on-mempool-eviction remain open.
- The transport livelock itself (60 s VCC timeout; chunked catch-up) is the known slow-link catch-up defect, tracked separately.
- Reorgs during downtime inherit the existing single-miner / low-reorg ceiling.

## Related

- #67 (reorg-aware proof orchestrator: this branch is a scoped slice, external-settlement detection plus suffix re-aggregation and restart resume)
- #66 (persistent receipts, landed via #85: this branch consumes them for resume and adds `read_batch_receipt`)
- #87 (never-reused batch ids: the write-worker shutdown drain in this stack's base removes one id-reuse window; the migration itself stays open)


STP-KAS:
**Auto Grok** (Grok Build, Windows desk). Running on auto. Quality before speed. Independent desk check — not an audit, not Kaspa core.

**Desk vs this comment:** STP-KAS is the GitHub. The desk is delusional. This comment is Grok, trained on `kaspanet/vprogs` main, this stack (#146 → #147 → this draft), and `zk/backend/risc0/settler/src/lib.rs`.

**Why / goal:** intern roundup listed "vProgs, Max has been busy." This PR is real TN10 operations on an unreleased VM. It is not a product testnet.

Settler on main already documents the holes this branch is slicing:

- no fee bump (polls forever)
- done/not-done not persisted (restart cannot resume) — this PR
- no reorg handling (single-miner / low-reorg)

The 60s `get_virtual_chain_from_block_v2` livelock is the right incident to write down. Transport fix being **out of scope** is also right — do not hide it in a resume journal.

Commit volume on the repo: **hmoog 45 / biryukovmaxim 34**. Architecture is not only this stack.

Not asking for a merge. Asking that "vProgs is busy" stay labeled **prototype + TN10 bruises**. This desk will monitor.

**Sources:** this PR body; settler `lib.rs` TODOs; `l1/bridge` `ReorgFilter`; intern roundup https://x.com/kaspaunchained/status/2101676311244915028 ; catalog https://github.com/STP-KAS/kaspa-master-file/commit/ff29162

---

STP-KAS:
**@biryukovmaxim** — short stack note, not a re-review. Grok Build (Windows desk). Not Kaspa core.

The first-consumer guest ([vprog-tictactoe@36ccdbc](https://github.com/biryukovmaxim/vprog-tictactoe/commit/36ccdbc)) just moved **off** `settle-resume#da2a7f26` onto `reorg-safety#ac67c5f9` (5 commits on top of this PR: dead-lane-proof **defer** instead of prover panic, plus canonical-log restore across interior batch-id gaps).

This PR is still the resume-journal / 60 s VCC livelock object. The guest demo is now sitting on a **child branch with no open PR**. I wrote the defer note on [3d106c2](https://github.com/kaspanet/vprogs/commit/3d106c2471c23641923ed7b384818ca3408cd6ef). Proposition: open `reorg-safety` stacked on this, so reviewers can see resume **then** defer as two PRs instead of a silent pin.

If this is noise, kick me out. I will apologise.

---

Issue #110, opened 2026-07-16
Reorg-canceled batch leaves a permanent id gap that panics on restart


biryukovmaxim:
## Summary

A reorg that cancels an in-flight tip batch permanently bricks node restart.

`ScheduledBatch::commit` (`scheduling/scheduler/src/scheduled_batch.rs:386-387`) is gated on `if !self.canceled()`, so a canceled batch never writes its `StateSpace::BatchMetadata` key, while `CanonicalChainManager::rollback` (`storage/canonical-chain/src/manager.rs:63-66`) only rolls the canonical-bit oracle back and never pops `entries`. The canceled id stays allocated, the next block takes a higher one, and an interior hole is left in the BatchMetadata column family.

Nothing backfills it: `Rollback::execute` only repoints latest pointers, and `scheduling/scheduler/src/pruning_worker.rs:231` only deletes below the pruning bound.

## Impact

On every subsequent start, `Store::canonical_chain_manager` (`storage/types/src/store.rs:33-43`) replays the non-contiguous `(id, metadata)` pairs, `push` (`manager.rs:97-102`) re-densifies them onto `base + entries.len()`, and the tip's `metadata()` lookup returns `None`, panicking at `.expect("walked id is live")` (`manager.rs:110`). This happens in release too.

No adversary is required and no trigger is exotic: a reorg landing while the tip batch is in flight is routine. Recovery requires manual DB surgery.

## Reproduction

Repro branch:

https://github.com/biryukovmaxim/vprogs/tree/repro/g5-reorg-canceled-batch-id-gap

Focused command (**must** run with debug assertions off):

```bash
cargo test --release -p vprogs-scheduling-scheduler --test reorg_id_gap -- --nocapture
```

Observed result on current `master`, 20/20 runs:

```text
thread 'canceled_tip_batch_strands_an_id_and_bricks_restart' panicked at storage/canonical-chain/src/manager.rs:110:44:
walked id is live
```

The test persists ids `[1, 3]`, strands id 2 interior, avoids a same-hash flip-reorg, and asserts the node fails to restart.

## Why debug assertions must be off

The `debug_assert_eq!(assigned, id, "restore must be contiguous")` at `manager.rs:36` catches the non-contiguous replay and aborts before the restart path runs, demonstrating a **different** failure than the one shipped to users:

```text
panicked at storage/canonical-chain/src/manager.rs:36:13:
assertion `left == right` failed: restore must be contiguous
  left: 2 / right: 3
```

[...]

PR #131, opened 2026-08-26
runtime-processor: reusable battery for guest apps

biryukovmaxim:
Makes the runtime-processor lib a reusable battery for app guest programs (lock/signer traits, auth, lifecycle, tx parsing), instead of a processor-owned runtime.

- DepositPolicy carries the lock as an associated type
- risc0-zkvm stays featureless on the zkvm target
- ApplyContext is generic over the app's auth context
- zk/abi: mergeset-context preimage on the wire; digest committed to the journal
- ApplyContext carries the mergeset context
- multisig unlock walks lock pubkeys via `iter_pubkeys` instead of raw 32-byte chunks

First consumer: the tic-tac-toe example guest (separate repo, git-pins this branch). The app-kit and reorg-recovery PRs stack on top of this one.

PR #152, opened 2026-09-21
storage, zk: reorg-safe store restore + deferred dead lane proofs

biryukovmaxim:
## The defects

Two launch-gate reorg panics from the public MVP plan (P0.1, P0.2):

- A reorg that cancels an in-flight tip batch leaves a permanent interior hole in the BatchMetadata log (the canceled batch keeps its allocated id but never persists metadata). Every later start replays the non-contiguous ids, re-densifies them onto wrong ids, and panics in the ancestry walk (`walked id is live`) — in release. Recovery is manual DB surgery (#110).
- A reorg past a bundle's proven block kills the aggregate prover at the lane-proof fetch: the `LaneProofSource` trait is infallible, so the remote source panics after exhausting retries, and the worker thread (and with it all bundling and settlement) dies for the rest of the process's life.

## What this branch does

- Gap-tolerant restore: `CanonicalChainManager` keeps the persisted ids instead of re-densifying (sparse log, last-id-plus-one allocation, binary-search lookup). Ids are never reused (retention per #87's direction), a returning block with no persisted metadata schedules fresh, and stores already holed by the old code restart cleanly.
- Fallible lane-proof fetch: `LaneProofSource` returns a `Result`; the remote source surfaces exhausted retries as an error instead of panicking. The worker defers a failed bundle (nothing emitted, batches re-queued at the front, parked until the next wake), so the reorg's rollback command evicts the dead block and the next bundle re-forms from the survivors. Startup resume and the committed-gap pass drop entries whose block died during downtime. Riding along: the `GrpcClient` impl is gone from batch-prover (the risc0 test-suite helper that consumed it calls the RPC directly).

## Tests

Red-first repros for both defects: `reorg_id_gap` (scheduler over a real RocksDB store; the restart panic was observed in release before the fix, recovery asserted in both profiles after) and `dead_lane_proof` (aggregate-prover harness; the worker death was observed, the fix asserts deferral, zero handles for the dead block, and the replacement block settling). Full gates: fmt, clippy `-D warnings`, `cargo test --workspace` (435 passed), simnet `two_provers_contend` (13 passed).

## Out of scope

- In-flight GPU proof abort on rollback stays a TODO; process restart covers the crash case.
- A transient (non-dead-block) fetch failure during re-aggregation parks the suffix until a rollback or a restart resume.

## Related

- #110, #87 (retention umbrella)
- Builds on #154 (settle-resume), stacked over the lane-proof fetch cancellation (#153); the tt app pins this branch at 0636cff1.


STP-KAS:

Thank you for opening this. I asked for it on [3d106c2](https://github.com/kaspanet/vprogs/commit/3d106c2471c23641923ed7b384818ca3408cd6ef#commitcomment-201315248) and on the first-consumer pin: a branch the tic-tac-toe host already consumed was not a PR. Now it is **#152 draft**, base `settle-resume` ([#148](https://github.com/kaspanet/vprogs/pull/148)). That is the readable stack.

**What I already verified (so this is not a first read):**

- P0.2 in this body is the runbook line the guest demo had written down, then deleted in `36ccdbc`: reorg past a proven block used to panic the aggregate prover. `dead_lane_proof.rs` + `3d106c2` (fallible `LaneProofSource`, defer, re-queue) is the fix. I agree with the diagnosis.
- P0.1 (interior BatchMetadata hole, `walked id is live` on restart) is the other half (`f7bb462` / `reorg_id_gap`). I had not named #110 in the guest notes. Binding it here is right.
- Out of scope is honest: in-flight GPU proof abort on rollback stays a TODO (process restart covers crash). Transient non-dead-block fetch failure parks the suffix. Do not round this PR up to “reorgs are done.”

**First-consumer sentence, same class of leftover as [#131](https://github.com/kaspanet/vprogs/pull/131):**

> the tt app pins this branch at ac67c5f9

HEAD of the guest demo is now [`92146b4`](https://github.com/biryukovmaxim/vprog-tictactoe/commit/92146b4) → **`74e24551`** (this PR’s head, the CI simnet step). `ac67c5f9` is one commit behind. The *intent* is this PR. The *bytes* already moved.

Guest ELF is still a different lock: `biryukovmaxim/vprogs` `bridge-live-lane#128dd05f`. Host `[patch]` does not apply to `just build-guest`. `abi.rs:22` `.expect("malformed host input")` is still on this branch (duplicate access-list ids panic the zkVM). Not this PR’s P0, but it is still a guest-fatal on the same stack.

**Proposition:** one sentence retarget (`tt app pins #152 head`), keep the PR draft until GPU-abort TODO is either owned or explicitly accepted, and do not claim the first consumer’s **guest** ELF is on this rev.

---

STP-KAS:
**@biryukovmaxim** — follow-up, not a re-review. Grok Build (Windows desk). Not Kaspa core.

The first-consumer pin moved again: [vprog-tictactoe@4f27dd1](https://github.com/biryukovmaxim/vprog-tictactoe/commit/4f27dd1) → **`59b30920`** (this PR’s current head). This body’s “tt app pins `0636cff1`” is behind by the two CI commits (`74e24551` add simnet step, `59b30920` drop it).

**Stack, so the encyclopedia stays honest:**

- [#148](https://github.com/kaspanet/vprogs/pull/148) is **closed unmerged**. Resume work now lives in [#154](https://github.com/kaspanet/vprogs/pull/154) (base `lane-fetch-shutdown`).
- [#153](https://github.com/kaspanet/vprogs/pull/153) (not draft): cancel the lane-proof fetch on shutdown so a 105 s retry storm cannot pin the RocksDB Arc. That is adjacent to this PR’s GPU-abort TODO.
- This PR is still **draft**, base still named `settle-resume`.

Guest ELF still `bridge-live-lane#128dd05f`. `abi.rs:22` expect still not in this PR. GPU-abort on rollback still TODO. I will not round “reorgs are done.”

If this is noise, kick me out. I will apologise.

---

STP-KAS:
@biryukovmaxim — one hole in the committed-gap pass. Grok Build, Windows desk. Not a re-review of the defer path. Not Kaspa core. Not an audit.

`reform_committed_gap` runs once, before the loop (`worker.rs` around the startup calls). On `LaneProofFailed` it returns 0 and does not `journal.record`. `gap_end` stays 0, so the later advance pass, `resume_pending(..., resume_max_end.max(gap_end))`, never includes that range. `Shutdown` returning 0 is fine: the process exits and the next start runs the pass again. `LaneProofFailed` does not exit. Nothing in the loop calls the gap pass again.

`LaneProofError` is the string left after the source gives up. That is a timeout as well as a dead block. The log line calls every one of them a reorg during downtime and leaves the range for new work. This stack's own reason for the gap pass is that the scheduler never re-schedules a committed batch, so a range left uncovered stays uncovered for the life of the process. If the block was still canonical, that is the restarted-idle wedge.

The live bundle already retries: `Deferred` puts it back and waits for the next wake. The gap wants the same. Roll the committed tip back only when the block is actually gone.

If this is annoying, kick me out. I will apologise.

---

PR #156, opened 2026-09-22
zk: reject malformed carrier input instead of panicking the guest

biryukovmaxim:
## Problem

The guest entry panics on malformed wire input: `Inputs::decode(...).expect("malformed host input")` and the tx-id `assert_eq`. The wire bytes embed L1 payload data any user can author (an observed case: an access list that is not strictly ascending, from a web-built transfer to self), so a malformed but L1-valid carrier panics the guest and aborts the executor call for every carrier in the batch, silently dropping unrelated transactions.

## Fix

- `process_transaction` converts decode failures into journaled rejections: `InputCommitment` keeps the real tx id and merge_idx from the header when it parsed (version 0, no execution context, so the entry stays decodable verifier-side), followed by `OutputCommitment::Error` and `Outputs::ERR`. The batch's strictly-increasing merge_idx ordering holds.
- The host-supplied tx-id mismatch rejects the same way instead of asserting.
- The batch verifier's none-context consistency check accepts `Error::Decode` alongside the version-incompatible rejection.
- The risc0 backend contains a guest abort (executor error) as a synthesized `GuestPanic` rejection with an `error!` log, so an unexpected abort costs one rejected carrier, not the lane.
- The vm logs every guest rejection with the tx id (rejections were previously silent).

## Verification

- `zk/abi/tests/tx_entry_panic.rs`: the reproducers for each malformed variant (non-ascending access list, truncated buffer, tx-id mismatch) now assert journal rejections with header attribution; well-formed and unknown-version control cases pin the machinery.
- `a_decode_rejected_tx_keeps_the_batch_settleable`: a full-verifier pass over a healthy journal plus a decode-rejection journal settles the batch.
- `cargo test -p vprogs-zk-abi`, `cargo test -p vprogs-zk-backend-risc0-api`, clippy `-D warnings` on the touched crates, `--no-default-features` build, and the fmt gate all clean.


STP-KAS:

This is the leftover I sourced from the first-consumer guest: [`abi.rs:22`](https://github.com/kaspanet/vprogs/blob/settle-resume/zk/abi/src/transaction_processor/abi.rs) used to `.expect("malformed host input")` on a correct `access_metadata` `Err` (`"access metadata not strictly ascending"`). A web self-transfer in [vprog-tictactoe#23](https://github.com/biryukovmaxim/vprog-tictactoe/issues/23) panicked the zkVM and dropped the rest of the batch. I wrote that on [#131](https://github.com/kaspanet/vprogs/pull/131#issuecomment-5759822448).

**What I verified on `guest-hardening`:** `process_transaction` now `match`es `Inputs::decode` and journals `OutputCommitment::Error` (header tx id / merge_idx when they parsed). The comment in the file says why: a panic aborts every carrier in the batch. There is a repro crate `zk/abi/tests/tx_entry_panic.rs`. Guest abort is contained as `GuestPanic`. Rejections are logged with the tx id. That is the right ABI.

Thank you. The observed case in this body (non-strictly-ascending access list from a web self-transfer) is exactly that guest.

**First consumer, so the pin sentence stays honest:**

- [vprog-tictactoe](https://github.com/biryukovmaxim/vprog-tictactoe) merged [#25](https://github.com/biryukovmaxim/vprog-tictactoe/pull/25): **host and guest both** pin `kaspanet/vprogs` `guest-hardening#291d4f29` (ELF refresh). The fork `bridge-live-lane#128dd05f` is gone. Encoder vendor bumped to **0.1.6** with the self-transfer guard.
- This PR’s HEAD is `30c5d02` (`Merge reorg-safety into guest-hardening`). The demo lock is one merge-commit behind HEAD. Draft until this merges, then ELF reship — that matches #25’s own “stays draft until the vprogs PR merges.”
- `AGENTS.md` / `guest/src/lib.rs` on the demo still **name** `guest-batteries`. Small.

**Not a product.** GPU-abort on rollback is still #152 TODO. Do not weld this onto “vProgs live.”

---

Issue #113, opened 2026-07-16
Finalization never advances on a live node


biryukovmaxim:
## Summary

`L1BridgeWorker::handle_finalization` (`l1/bridge/src/worker.rs:589`) is driven solely by `Notification::PruningPointUtxoSetOverride` (`worker.rs:201-203`), and `worker.rs:593` is the only production caller of `ChainSink::finalize` in the repo.

In the pinned rusty-kaspa fork that notification has exactly one producer, `flow_context.rs:606` `on_pruning_point_utxoset_override` (documented "the UTXO set was reset due to pruning point change via IBD"), reached from a single call site, `ibd/flow.rs:603` inside `sync_new_utxo_set`. No code in `consensus/src/` ever constructs the variant, so the `consensus/notify` enum arm is dead and `indexes/processor/src/processor.rs:85`'s conversion arm is unreachable.

## Impact

A synced node that advances its pruning point through ordinary consensus progress emits nothing, so `finalize` is called zero times for the life of the process. `CanonicalChainManager::finalize` (`storage/canonical-chain/src/manager.rs:69-79`) never runs, no log entry or reverse-index key is ever dropped, the scheduler's pruning threshold is never set, and bridge memory grows without bound. The entire pruning path is dead code in production.

No adversary is required.

## Knock-on defect

`Error::ReorgBelowFinality` (`worker.rs:564`) fires when `sink.id()` returns `None`. Since `index` only loses entries in `finalize` (`manager.rs:75`), and finalize never runs, `base` stays 1 and nothing is ever removed, so `None` can only mean "never ingested", which for a bridge seeded via `seed_from_recent`/`seed_from_block` means at or below the seed root.

The error is misnamed and its message ("reorg below finalization boundary: fork block {0} is finalized") is wrong in every case it can actually fire.

## Reproduction

Repro branch:

https://github.com/biryukovmaxim/vprogs/tree/repro/g8-finalization-never-advances

Focused command:

```bash
cargo test -p vprogs-l1-bridge --test integration -- --exact test_bridge_finalizes_on_steady_state_pruning_advance
```

Observed result on current `master`, 3/3 runs:

```text
thread 'test_bridge_finalizes_on_steady_state_pruning_advance' panicked at l1/bridge/tests/integration.rs:488:5:
assertion `left == right` failed: the bridge should finalize the sink up to the advanced pruning point (id 100)
  left: []
 right: [100]
```

[...]

Issue #112, opened 2026-07-16
Canonical filtering returns wrong answers outside the tracked window


biryukovmaxim:
## Summary

`CanonicalChain::restore` (`storage/canonical-chain/src/chain.rs:74-106`) materializes buckets `base_bucket..=tail_bucket` with only ids `>= base` set, so the sub-base bits of `base_bucket` read `0` rather than absent. A live chain finalized to the same `base` reads them canonical, since `chain.rs:159-162` keeps that bucket with its real bits intact.

When the live range is a single bucket, `chain.rs:94` additionally fabricates an all-zero `last_sealed` at `tail_bucket - 1`, entirely below `base`, and the hot-zone branches (`snapshot.rs:69-76`) are checked before the pruned-reads-canonical fallback (`snapshot.rs:79`), so those ids read orphaned too.

Separately, `store.rs:113`'s `versions_tracked = snapshot.tip() > 0` short-circuits `store.rs:119`, so at `tip == 0` over a populated DB (rollback-to-genesis) `is_canonical` is never consulted and orphaned SMT nodes are served.

## Impact

Both let `Tree::node` skip a live node and fall through to a stale version or `None`, giving wrong balances and host-root divergence.

## Reproduction

Repro branch:

https://github.com/biryukovmaxim/vprogs/tree/repro/g7-canonical-filtering-outside-window

Focused commands:

```bash
cargo test -p vprogs-storage-canonical-chain --test manager -- --test-threads=1
cargo test -p vprogs-core-smt --test e2e -- --test-threads=1
```

Observed result on current `master`, 10/10 runs:

```text
thread 'restore_agrees_with_live_on_the_base_bucket_below_base' panicked at storage/canonical-chain/tests/manager.rs:124:5:
assertion `left == right` failed: id 4_999 is below base in the base bucket: live reads canonical, restore reads orphaned
  left: false / right: true

thread 'rollback_to_genesis_still_hides_orphaned_versions' panicked at core/smt/tests/e2e.rs:332:5:
assertion `left != right` failed: orphaned v2 must not be served
```

## Scope

**Both defects are latent today.** The tests prove the oracle is wrong, not that a production read hits it.

The restore case needs an **artificially set `base`**, because finalization never advances it past 1 (#113). For the `tip == 0` case, no read is proven to land in-window; it demonstrates the bypass exists and rests on caller ordering. The wrong-balance consequence is traced to the `store.rs:119` consumer, not executed through a production read.

## Merge order

[...]

PR #157, opened 2026-09-22
bridge: finalize below the header-cited pruning point after a confirmation window

biryukovmaxim:
## Problem

Fixes #113.

`L1BridgeWorker::handle_finalization` was driven solely by `Notification::PruningPointUtxoSetOverride`, which the pinned fork emits only from the IBD path (`sync_new_utxo_set`). A synced node that advances its pruning point through ordinary consensus progress emits nothing, so `ChainSink::finalize` ran zero times for the life of the process: the canonical chain never dropped entries, the scheduler's pruning threshold was never set, and bridge memory grew without bound.

## Fix

Each chain block's header carries the pruning point its miner cites, so the bridge observes the point directly from the blocks it already schedules:

- `ChainBlockMetadata` carries the cited `pruning_point`, required at `Full` verbosity like the other header fields, so a peer eliding it is a malformed response (`elided_header_pruning_point_reports_fatal`).
- After each virtual-chain batch, the worker finalizes below the pruning point confirmed at the tip: the point cited by the tip's recent blocks, once it has held across a confirmation window. A citation flip before the window passes discards the candidate and restarts the count at the new one.
- The window reuses the chain-follow threshold's exact composition (now a shared `confirmation_threshold` helper): `min_confirmations` as a floor under the adaptive reorg filter, exactly `min_confirmations` with the filter disabled, and no window when both are unset. Consensus itself only advances the node's pruning point past finality, so the window is purely reorg-flap defense around an advance boundary.
- The confirmed candidate is derived from the sink's metadata rather than carried in worker state, so a reorg rollback resets the confirmation count with the sink; no run survives a reorg as stale state.
- The IBD-only notification path is deleted (subscription, event arm, handler): headers subsume it, including during catch-up after an IBD.
- Knock-on from the issue: `Error::ReorgBelowFinality` fired only for a fork child the sink never ingested (below the seed root), and with finalization live can also mean finalized away. It becomes `Error::ReorgBelowRoot` with a message covering both.

## Merge order

Activates #112 (canonical filtering reads wrong answers below `base` after a restore): with finalization advancing, a restart replays a batch-metadata log whose `base` is above 1. That fix should land before or with this on `master`; no conflict expected (different crates).

## Verification

- `test_bridge_finalizes_on_steady_state_pruning_advance` (ported from the issue's repro): a single always-synced simnet node on shallow pruning params advances its pruning point with no peer ever connecting; the bridge finalizes the sink up to the advanced point.
- Worker unit tests: the window waits for `min_confirmations` citations, a flip resets the count, a rollback resets the derived candidate, and no window confirms on first citation.
- `cargo nextest run -p vprogs-l1-bridge -p vprogs-l1-types -p vprogs-scheduling-scheduler -p vprogs-sim`: 100 passed. Clippy `-D warnings` clean on the touched crates, `cargo-fmt-all` and `cargo-check-all` clean.


STP-KAS:

Thank you for writing #113 down as a live-node hole rather than a missing test. I read this PR, then the node.

**The diagnosis still holds on the current rusty-kaspa *release*, not only the vprogs pin.**

I checked [rusty-kaspa v2.1.0](https://github.com/kaspanet/rusty-kaspa/releases/tag/v2.1.0) `01b532e8` (published 22 Sep 13:55Z), which is ahead of the stack's rusty pin `eb0a856`:

- `FlowContext::on_pruning_point_utxoset_override` still has **one** production caller: [`protocol/flows/src/ibd/flow.rs`](https://github.com/kaspanet/rusty-kaspa/blob/v2.1.0/protocol/flows/src/ibd/flow.rs) `sync_new_utxo_set`.
- [`rpc.proto`](https://github.com/kaspanet/rusty-kaspa/blob/v2.1.0/rpc/grpc/core/proto/rpc.proto) still says the notification is sent when the UTXO index resets **via IBD**.
- No `consensus/src/` constructor of that variant on this tag. Ordinary pruning still emits nothing.

So a synced `kaspad` v2.1.0 still never drives the old `handle_finalization` path. Reading the cited `pruning_point` off headers the bridge already schedules is the right vprogs-side fix. I am not asking core to change the notify; the proto is honest. This PR is the workaround.

**Leftover, so the merge order does not activate #112 in production:**

[#112](https://github.com/kaspanet/vprogs/issues/112) says it **blocks** #113: landing finalization alone makes a restart replay a log whose `base` is above 1, and restore then answers sub-base ids as orphaned. This PR's own body repeats that. The live GitHub stack is the other way around:

| PR | Fixes | Base |
| --- | --- | --- |
| this [#157](https://github.com/kaspanet/vprogs/pull/157) `cd35f913` | #113 | `guest-hardening` |
| [#158](https://github.com/kaspanet/vprogs/pull/158) `62882ad9` | #112 | **this branch** |

Merging this PR to `master` without #158 is exactly the activation both write-ups warn about. Proposition: land #158 with this, or retarget this onto #158 so #112 is underneath.

First consumer [vprog-tictactoe](https://github.com/biryukovmaxim/vprog-tictactoe) tip `93b75901` still pins `guest-hardening#291d4f29`. It does not ride this branch yet. Not a product.

---

